PR #34, opened 2017-02-26
Fix Boolean Logic and Add gunicorn Config

ChristopherHammond13:
As described.

I messed up the variable that checks whether we're in production or not so actually wrote a strtobool function. Yay.

Also added a config file for gunicorn.


ChristopherHammond13:
Getting bored of waiting :P Going to merge and run in prod so that we can launch tomorrow. There may well be a cleaner way of doing .env boolean comparisons, but we can do that in a future release. For now, I must gunicorn!

PR #41, opened 2017-02-28
sort apps by creation date

hennersz:


jermenkoo:
Looks good to me.

xoneco:
does still need build files?

hennersz:
I wasn't sure if having different built files on separate branches would cause loads of merge conflicts so i didn't commit them

hennersz:
so this will need to be rebuilt on the server

xoneco:
@hennersz so we merge the other one first and then add build files to this (maybe with a rebase)

OR @ChristopherHammond13 has to run the build on the box.
Chris do you know the steps?

Ideally we should be consistent tho. So either no build files in version control, or all build files in version control

ChristopherHammond13:
@xoneco do I just need to do a `collectstatic` once I git pull, or is there more to this?

hennersz:
tbh id rather no build files in version control
@ChristopherHammond13 wait until I create the production build scripts and I will add instructions

ChristopherHammond13:
@hennersz It would be ideal to not have to do static site generation on the servers otherwise we'll need to install bundler / ruby /etc.

xoneco:
this is more of a #future #biggerpicture opinion for later down the line, but imho all builds for prod should happen as part of the build/deployment pipeline, not locally.

hennersz:
also you only need node to build the frontend

hennersz:
but for now I will just build production code to a separate directory

hennersz:
actually this is going to make everything more complicated

PR #48, opened 2017-03-01
object_hook to convert datetime string to the object while obtaining …

abdulfaizp:
…it from the database

@jermenkoo @xoneco I couldn't reproduce the sentry 500 with the same query. So I am guessing the error is because datetimes dont get decoded in to a datetime format when i take query params from the database. hence the [object hook](http://stackoverflow.com/a/16310732/5556675).

### DIFF
--- a/backend/uclapi/roombookings/models.py
+++ b/backend/uclapi/roombookings/models.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 from django.db import models
 from .api_helpers import generate_token
 import json
+import datetime
 
 
 class Booking(models.Model):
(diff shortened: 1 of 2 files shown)


xoneco:
As mentioned yesterday, could you write some tests for this one pls :)

ChristopherHammond13:
Can this be merged without a test, @xoneco ?

xoneco:
No, we're not really sure what this does or which problem it solves. This one is a bit stale

@abdulfaizp ?

jermenkoo:
@xoneco it should fix the following issue: https://sentry.io/isd-api-team/api/issues/227162037/

abdulfaizp:
@xoneco this code is purely based on what sentry has reported. it will solve the exact issue sentry sent us. But it seems to be the case sometimes databse accepts stringified datetime. sometimes it does not. i am not sure why :/

xoneco:
Hmm ok, so we should not do anything about this for now. Let's leave the branch alive and and if the error occurs again or if we can reproduce it, then we can deploy the fix.

abdulfaizp:
this wont be relevant once we swal django ORM with cx_Oracle. so we can close this

PR #51, opened 2017-03-04
add `ok` field to token-auth-fails

abdulfaizp:
`ok` field added to decorator response

### DIFF
--- a/backend/uclapi/roombookings/token_auth.py
+++ b/backend/uclapi/roombookings/token_auth.py
@@ -1,4 +1,3 @@
-from rest_framework.response import Response
 from dashboard.models import App
 from django.core.exceptions import ObjectDoesNotExist
 from django.http import JsonResponse
@@ -9,18 +8,22 @@ def does_token_exist(view_func):
         token = request.GET.get("token")
 
         if not token:
-            return JsonResponse({
+            response = JsonResponse({
                 "ok": False,
                 "error": "No token provided"
             })
+            response.status_code = 400
+            return response
 
         try:
             App.objects.get(api_token=token)
         except ObjectDoesNotExist:
-            return JsonResponse({
+            response = JsonResponse({
                 "ok": False,
                 "error": "Token does not exist"
             })
+            response.status_code = 400
+            return response
 
         return view_func(request, *args, **kwargs)
     return wrapped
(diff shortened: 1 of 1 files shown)


xoneco:
Just to post this somewhere offline:

- Still needs response `Bad Request` in response type.

tiferrei:
Just for the record: Partially fixes issue #52

backend/uclapi/roombookings/token_auth.py
```diff
@@ -9,18 +8,22 @@ def wrapped(request, *args, **kwargs):
         token = request.GET.get("token")
 
         if not token:
-            return JsonResponse({
+            response = JsonResponse({
                 "ok": False,
                 "error": "No token provided"
             })
+            response.status_code = 400
```
tiferrei: I think this should a `401` 👉  [Internet Engineering Task Force (IETF)](https://tools.ietf.org/html/rfc7235#section-3.1)

xoneco:
This should be good to go.

@jermenkoo / @hennersz could you CR and merge, pls?

backend/uclapi/roombookings/token_auth.py
```diff
@@ -9,16 +8,22 @@ def wrapped(request, *args, **kwargs):
...
 
         try:
             App.objects.get(api_token=token)
         except ObjectDoesNotExist:
-            return JsonResponse({
+            response = JsonResponse({
+                "ok": False,
                 "error": "Token does not exist"
             })
+            response.status_code = 400
```
tiferrei: I think this should a `401` 👉  [Internet Engineering Task Force (IETF)](https://tools.ietf.org/html/rfc7235#section-3.1)

xoneco:
The returning JSON basically takes over the function of being descriptive. Just reading through that section you linked, we don't fulfill like the majority of things they mention (Authorization header, WWW-Authenticate, etc.)

We also use `400` in various other parts of the API, so we should keep it like that for now for consistency. I'm open to changing this API-wide, but that should be done in comprehensive way in a separate PR.

ChristopherHammond13:
Are we good to merge this? If so I'll get it straight into prod now. My only concern is if this constitutes a _breaking_ change; if so, we ought to version this change. @xoneco @abdulfaizp

xoneco:
No, this is not a breaking change. It changes the API to behave the way it is advertised

ChristopherHammond13:
Merged :)

PR #54, opened 2017-03-10
Integrate Keen for better metrics

xoneco:
Keen is another tool that makes metrics a bit simpler and is tied around specific events we send them. I use it for Simple Poll and it's super useful. https://keen.io

Pls CR @abdulfaizp @jermenkoo 

@ChristopherHammond13 before you merge this we need to create two environment variables on both the staging and prod boxes


backend/uclapi/roombookings/decorators.py
```diff
@@ -0,0 +1,63 @@
...
+            "service": service,
+            "method": method,
+            "version-headers": {},
+            "queryparams": queryparams
+        }
+
+        if len(version_headers) > 0:
+            parameters["version-headers"] = version_headers
+
+        # import pdb; pdb.set_trace()
```
jermenkoo: maybe nuke this comment away?

ChristopherHammond13:
@xoneco I'm ready to integrate this now! I'll get it into staging then when we have Shibboleth activated there I'll pull down to master on prod.

PR #60, opened 2017-03-17
Add coordinate data to rooms output

xoneco:
This gets coordinate date which exists within Estates, but not within roombookings (CMIS).
We store this in our own database using the new management command in this PR.

UPDATE: Thanks to @anirudhpillai we have lots of valid data for this now.   
Just waiting for https://github.com/uclapi/uclapi/pull/66 to get merged, because this PR depends on it.

Related docs PR:
https://github.com/uclapi/apiDocs/pull/12

Change notification:
https://twitter.com/uclapi/status/846693147107344385


abdulfaizp:
lgtm. but storing it this way would mean we will hit database once for every rooms. which we dont really have to since there are only 300 rooms. if we store the entire data in one column, its only one hit and not 300 hits for every requests for rooms.

backend/uclapi/roombookings/helpers.py
```diff
@@ -112,23 +112,48 @@ def _parse_datetime(start_time, end_time, search_date):
...
+                    "address": [
+                        room.address1,
+                        room.address2,
+                        room.address3,
+                        room.address4
+                    ]
+                }
+            })
+        else:
+            rooms.append({
```
abdulfaizp: this repetition can be avoided by putting the entire insertion of coordiantes in a try except block. it will look neater and its concise.

xoneco:
@abdulfaizp django makes one connection to the db per request, so it's not as bad as it looks. Storing it all in one field just seems ugly? I'll try and do some performance testing.

xoneco:
travis

abdulfaizp:
@xoneco this

backend/uclapi/roombookings/management/commands/update_locations.py
```diff
@@ -0,0 +1,34 @@
...
+                    loc.save()
+                except ObjectDoesNotExist:
+                    loc = Location(
+                        siteid=siteid,
+                        lat=lat,
+                        lng=lng
+                    )
+                    loc.save()
+            except:
+                pass
```
xoneco: what is this?

abdulfaizp: no location for that room.

xoneco: Every room has coordinates though

xoneco:
leaving WIP on this until https://github.com/uclapi/uclapi/pull/66 is merged.

PR #69, opened 2017-03-28
Remove files needed for initial insert

xoneco:
## What does this PR do?
Remove files we needed initially for the coordinate insert, but longer require.
Hence, cleanup for clarity and cleanliness.

## ✅ Pull Request checklist

- [x] Is this code complete?
n/a Are tests done/modified?

n/a

No

## :squirrel: Deploy notes
n/a


ChristopherHammond13:
We need to discuss the best way of dynamically generating and gathering initial data sets. We should be able to deploy a production environment from scratch with ease :)

xoneco:
As discussed in person:

- The single source of truth should be the db, not a JSON file
- If we make any updates to this, it will happen via Django Admin or something similar, not via another JSON file
- We **do** need to think about how to replicate this sort of constant data from one environment to another when setting it up initially, but that solution is not blocked by this PR

xoneco:
Crowdsourcing would be very cool!
That's potentially a whole nother project with a huge scope 😄 

But would be super exciting and I want to do it 🌮

abdulfaizp:
I think a lot of the problems in UCL can be solved by crowdsourcing a lot of stuff. Things like room location, free space, water fountains etc.

ChristopherHammond13:
@xoneco @abdulfaizp did we come to a conclusion on this? Shall I merge?

xoneco:
@ChristopherHammond13 This can be merged, yes. But we need some kind of db/table replication mechanism.

ChristopherHammond13:
@xoneco we already have replication on the production database via Amazon RDS (Multi-AZ deployment; we essentially have a replica on standby in another availability zone ready for a switchover if the main deployment fails). Is this enough? We could have a replica setup if you're concerned but I think this is enough.

PR #81, opened 2017-06-01
using Redis as caching backend for throttling

jermenkoo:
.## What does this PR do?
This PR finally implements rate-limiting as desired.

Previously the solution allowed high limits as `gunicorn` has multiple workers (i.e. Django instances) and throttling data would be stored in memory. We solve this by using Redis to store the throttling data, meaning each worker will be always informed about the real number of requests client has made so far.

## ✅ Pull Request checklist

- [x] Is this code complete?

No.

No.

## :squirrel: Deploy notes
A new AWS ElastiCache Redis instance (`cache.t2.micro` should be sufficient) needs to be created.

`.env` file needs to be updated as well, followingly:
The `REDIS_UCLAPI_HOST` should be the IP of your Redis instance.

## Anything else
No.


wilhelmklopp:
I'm thinking we should start separating our requirement files into `deploy-requirements.txt`  `development-requirements.txt` `test-requirements.txt` and so on to have a clear separation for what dependencies are needed where.

Thoughts @ChristopherHammond13 @jermenkoo ?

jermenkoo:
@wilhelmklopp I definitely agree, as now we use multiple modules `fakeredis`, `freezegun` for the sake of testing. I found [this answer on StackOverflow](https://stackoverflow.com/questions/17803829/how-to-customize-a-requirements-txt-for-multiple-environments) and it might be good to follow it.

wilhelmklopp:
@jermenkoo oh, that looks good! +1 for following that.

ChristopherHammond13:
Redis should be fine, how about if we have distributed redis with each web host acting as a redis host? Is that totally stupid, or is it a good use of resources?

jermenkoo:
@ChristopherHammond13 Amazon ElastiCache (Redis) supports clustering as well, preventing single point of failure.

@wilhelmklopp changed `.env.example` as requested.

ChristopherHammond13:
@jermenkoo wasn't too bothered about the clustering aspect (because this is only for rate limiting, so redis being down shouldn't bring the API down), so much as I was bothered about how many hosts we were going to provision :P

wilhelmklopp:
+1 for clustering because no single point of failure
-1 for putting this on ec2 because separation of concerns :P

@jermenkoo how does this change the local set up by the way?
We should update local setup instructions in the knowledge base

wilhelmklopp:
Redis will be very useful once we do async processing using something like celery :)

backend/uclapi/uclapi/settings.py
```diff
@@ -11,10 +11,8 @@
 """
 
 import os
-import raven
 import requests
 from .utils import strtobool
-from .dbrouters import ModelRouter
```
ChristopherHammond13: haha, well caught unnecessary include!

backend/uclapi/.env.example
```diff
@@ -64,9 +64,13 @@ OPBEAT_ORG_ID=
...
+# Keen settings
 KEEN_PROJECT_ID=
 KEEN_WRITE_KEY=
 
 # Sentry settings
 SENTRY_DSN=
+
+# Redis settings
+REDIS_UCLAPI_HOST=
+REDIS_UCLAPI_CLIENT_CLASS=
```
ChristopherHammond13: Could you explain what the client_class variable does, and maybe include an example in the .env.example?

backend/uclapi/requirements.txt
```diff
@@ -8,7 +8,9 @@ cx-Oracle==5.3
 Django==1.11.1
 django-cors-headers==2.0.2
 django-dotenv==1.4.1
+django-redis==4.8.0
 djangorestframework==3.6.3
+fakeredis==0.8.2
```
ChristopherHammond13: How can fakeredis be used, as it doesn't appear to be referenced anywhere else

jermenkoo:
In `backend/uclapi/uclapi/settings.py`, there is a new `CACHES` variable, which defines the available caches to Django. There are 2 entries:

* `default` - defaults to the [cache Django uses originally](https://docs.djangoproject.com/en/1.11/topics/cache/#local-memory-caching)

* `uclapi` - our Redis caching backend. 
    `CLIENT_CLASS` defines how `django_redis` should talk to the Redis instance  (we do not want to change this). `REDIS_CLIENT_CLASS` defines what the Redis instance at `REDIS_UCLAPI_HOST` is.

`fakeredis` is not used directly, but we need it so that during tests we can mock the lack of a real Redis instance to test it against.

backend/uclapi/requirements.txt
jermenkoo: It is used for dev environment so that you can dev against it & your tests pass without needing a real Redis instance.

ChristopherHammond13:
It works great now! Squashing and merging. Great work :)

PR #82, opened 2017-06-05
Management Commands for Caching Equipment and Rooms

ChristopherHammond13:
## What does this PR do?
Adds caching for Equipment and Rooms

## ✅ Pull Request checklist

- [ ] Is this code complete?

No

## :squirrel: Deploy notes
Need to migrate `gencache` then immediately update the cache on all three endpoints


ChristopherHammond13:
Changes that need to be done:
1) I'd like the cache update command to do all three endpoints back to back if possible so that it can then switch bucket.
2) The field sizes are all too large. They should be set to the sizes of the fields in the Oracle database to save memory: remember, this is all going into RAM.

backend/uclapi/roombookings/management/commands/create_rooms.py
```diff
@@ -0,0 +1,43 @@
...
+                    bookabletype=room[10],
+                    roomclass=room[11],
+                    zone=room[12],
+                    webview=room[13],
+                    automated=room[14],
+                    capacity=room[15],
+                    category=room[16]
+                )
+                room_cache.save()
+            except IndexError:
```
wilhelmklopp: this seems like it will cause trouble

Faiz: why?

wilhelmklopp: There shouldn't ever be an index error. If there is one, something is wrong.
This suppresses the warning without fixing the cause, right?

jermenkoo: L18 already starts from the 2nd (index 1) element, why do we need this then?

backend/uclapi/roombookings/management/commands/create_rooms.py
```diff
@@ -0,0 +1,43 @@
+from django.core.management.base import BaseCommand
+from roombookings.models import RoomCache
+
+
+class Command(BaseCommand):
+
+    help = 'Clone room data to gencache'
+
+    def handle(self, *args, **options):
+        with open('roombookings/management/commands/data/rooms.csv') as f:
```
wilhelmklopp: what is this file?

ChristopherHammond13:
Closing this as we are unlikely to cache data in this way. We'll probably unify data caching for `roombookings` and `timetable` in the future, but as `equipment` and `rooms` loads quickly enough without a cache we probably won't do that.

PR #83, opened 2017-06-05
Implementing PostgreSQL Connection Pooling with Gunicorn Support

ChristopherHammond13:
## What does this PR do?
It will eventually replace all connections to Postgres with a connection pool. This should hopefully stop us hitting the `max_connections` variable that is defined in pg.
This value differs depending on whether the server is a `t2.micro` or a `t2.small`, etc., so we are using environment variables to configure this.

I hope that this will make the servers more reliable under high load. At the moment, a high load can cause dropped off connections due to Postgres refusing to service them. With this implementation requests would actually get serviced; they would just be slower (as you would expect under high load).

## ✅ Pull Request checklist

- [ ] Is this code complete?

No

## :squirrel: Deploy notes
Just restart the API after pulling the latest code. Don't forget to install the new requirements. The new gunicorn config should get read once the restart command is triggered.

## Anything else
This is still a WIP and advice is welcomed!

### DIFF
--- a/backend/uclapi/.env.example
+++ b/backend/uclapi/.env.example
@@ -31,6 +31,12 @@ DB_UCLAPI_PASSWORD=
 DB_UCLAPI_HOST=
 DB_UCLAPI_PORT=5432
 
+# The size of the PostgreSQL connection pool to use for the main db
+# Set this to some value slightly smaller than max_connections divided by the
+# number of EC2 hosts provisioned to host the API, divided by the number of
+# gunicorn workers
+DB_UCLAPI_POOL_SIZE=
+
 ### Oracle Room Bookings Settings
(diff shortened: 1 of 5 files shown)


wilhelmklopp:
Will this break anything/require changes for us running `uclapi` locally? - just checking

ChristopherHammond13:
@wilhelmklopp no changes to how we work locally I don't think; we'll just need to pip install when doing a git pull. The pooling is done by a library. It may actually make sense to start developing with gunicorn running. I need to do some testing because the problem wasn't solved when using the Django test server, but that's probably because the greenifying of psycopg2 is done in the `postFork` gunicorn function.

ChristopherHammond13:
I am closing this now because the work is going to be done by https://github.com/pgbouncer/pgbouncer which only requires `.env` changes

PR #84, opened 2017-06-05
OAuth Support

ChristopherHammond13:
## What does this PR do?
It is a work in progress implementation of OAuth built on top of UCL's Shibboleth SSO system.

It is currently a Work in Progress and fails to adhere to a lot of the OAuth standard at the moment, but now that we have redis it will be much easier to cache away codes, for example, in order to bring us closer to the OAuth 2.0: [RFC6749](https://tools.ietf.org/html/rfc6749).

The specification defined in this code at this time is both undocumented and subject to change, so developers should not yet attempt to build client libraries on top of this. When the code is complete we will release it into production and also announce a number of libraries written for you.

## ✅ Pull Request checklist

- [x] Is this code complete?

No

## :squirrel: Deploy notes
Install requirements, run migrations, ???, Profit.

## Anything else
As noted, this is currently a _work in progress_ and the specification is not yet adhered to. Suggestions are, of course, welcome, as are tests and code review.


backend/uclapi/.env.example
```diff
@@ -22,6 +22,13 @@ UCLAPI_DOMAIN=uclapi.com
 # See https://dryan.com/articles/elb-django-allowed-hosts/ for more info.
 UCLAPI_RUNNING_ON_AWS_ELB=False
 
+# Set URLs that may not be used as callback URLs (for security reasons).
+UCLAPI_OAUTH_CALLBACK_DENIED_URLS=localhost;127.0.0.1
```
wilhelmklopp: I assume this includes non-standard ports?

backend/uclapi/dashboard/api_applications.py
```diff
@@ -179,3 +191,191 @@ def delete_app(request):
...
+        })
+        response.status_code = 400
+        return response
+
+    
+    # Check if the new callback URL uses an acceptable protocol (e.g. HTTP or HTTPS)
+    # The list comprehension will explode the UCLAPI_OAUTH_CALLBACK_ALLOWED_PROTOCOLS environment
+    # variable, split by semicolons, append each one with :// then check if the URL starts with it.
+    # If none of them work then bail out.
+    if not any([new_callback_url.startswith(p + "://") for p in os.environ.get("UCLAPI_OAUTH_CALLBACK_ALLOWED_PROTOCOLS").split(';')]):
```
wilhelmklopp: split over multiple lines for PEP8 compliance

backend/uclapi/dashboard/migrations/0006_auto_20170516_2014.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
```
wilhelmklopp: Might need to do these migrations again based on what is on current master. If I understand migrations correctly?

backend/uclapi/dashboard/models.py
```diff
@@ -35,6 +36,30 @@ class App(models.Model):
...
+        default=generate_app_client_id
+    )
+
+    client_secret = models.CharField(
+        max_length=64,
+        unique=True,
+        default=generate_app_client_secret
+    )
+
+    callback_url = models.CharField(max_length=500, default="")
```
wilhelmklopp: You could use `URLField` here

backend/uclapi/dashboard/urls.py
```diff
@@ -9,5 +9,9 @@
     url(r'api/rename/$', rename_app),
     url(r'api/regen/$', regenerate_app_token),
     url(r'api/delete/$', delete_app),
+    url(r'api/setcallbackurl/$', set_callback_url),
```
wilhelmklopp: can we use snake_case or camelCase for these?

backend/uclapi/dashboard/urls.py
```diff
@@ -9,5 +9,9 @@
...
+    url(r'api/setscope/roombookings/$', set_rb_scope),
+    url(r'api/setscope/timetable/$', set_timetable_scope),
+    url(r'api/setscope/uclu/$', set_uclu_scope),
```
wilhelmklopp: I think these three setscope endpoints should all be the same with the name of the scope as a queryparam. Seems cleaner and better for scale.

backend/uclapi/dashboard/views.py
```diff
@@ -115,7 +115,17 @@ def dashboard(request):
...
-            "updated": app.last_updated
+            "updated": app.last_updated,
+            "oauth": {
+                "client_id": app.client_id,
+                "client_secret": app.client_secret,
+                "callback_url": app.callback_url,
+                "scope": {
+                    "private_roombookings": app.scope.private_roombookings,
+                    "private_timetable": app.scope.private_timetable,
+                    "private_uclu": app.scope.private_uclu
```
wilhelmklopp: Are these scope values just True or False?

backend/uclapi/oauth/models.py
```diff
@@ -0,0 +1,61 @@
+from django.db import models
+from .app_helpers import generate_user_token
+
+class OAuthScope(models.Model):
+    # We should really have a primary key for the OneToOneField
+    id = models.AutoField(primary_key=True)
```
wilhelmklopp: this is implicit and can be left out :) (for the future)

backend/uclapi/oauth/templates/permissions.html
```diff
@@ -0,0 +1,28 @@
...
+
+<body>
+
+	<script>
+    window.initialData = {{ initial_data|safe }};
+    window.staticURL = {% static "" %}
+  </script>
+	<div class="app"></div>
+  <div class="csrfDiv">
+  <script src="{% static "js/authorise.js" %}"></script>
```
wilhelmklopp: the name of the template and JavaScript file should probably have the same name.

backend/uclapi/oauth/views.py
```diff
@@ -0,0 +1,347 @@
...
+
+    try:
+        eppn = request.META['HTTP_EPPN']
+        groups = request.META['HTTP_UCLINTRANETGROUPS']
+        cn = request.META['HTTP_CN']
+        department = request.META['HTTP_DEPARTMENT']
+        given_name = request.META['HTTP_GIVENNAME']
+        display_name = request.META['HTTP_DISPLAYNAME']
+        employee_id = request.META['HTTP_EMPLOYEEID']
+    except:
```
wilhelmklopp: as mentioned, these should except specific exceptions.

backend/uclapi/oauth/views.py
```diff
@@ -0,0 +1,347 @@
...
+        return JsonResponse({
+            "ok": False,
+            "error": "The JSON data was not in the expected format. Please contact support."
+        })
+
+    app = App.objects.get(client_id=data["client_id"])
+    state = data["state"]
+
+    redir = app.callback_url + "/denied?state=" + state
```
wilhelmklopp: denied should be queryparam, not path

backend/uclapi/oauth/views.py
```diff
@@ -0,0 +1,347 @@
...
+    verification_data_str_enc = signer.sign(verification_data_str)
+
+    full_verification_data = {
+        "client_id": app.client_id,
+        "state": state,
+        "verification_data": verification_data_str_enc
+    }
+
+    try:
+        vr = requests.post(app.callback_url + "/verify", data=full_verification_data)
```
wilhelmklopp: As discussed and to be compliant with OAuth, no backend requests should be happening after user clicks allow. There should be a redirect straight to the callback after the user clicks allow.

backend/uclapi/oauth/views.py
```diff
@@ -0,0 +1,347 @@
...
+    # Just in case they've tried to be super clever and host multiple apps with the same
+    # callback URL, we'll provide the client ID along with the state
+    return redirect(app.callback_url + "?client_id=" + app.client_id + "&state=" + state)
+
+def userdata(request):
+    try:
+        token_code = request.GET.get("token")
+        token = OAuthToken.objects.get(token=token_code)
+    except:
+        return JsonResponse({
```
wilhelmklopp: I'm guessing this response should be a more general one that future apps/services can also make use of. Probably a decorator?

backend/uclapi/oauth/views.py
```diff
@@ -0,0 +1,347 @@
...
+    # Now forward them the OAuth token for the user. If they're not happy with that then
+    # that's their fault after the final redirect!
+    oauth_req = requests.post(app.callback_url + "/token", data=oauth_data)
+
+    # Now redirect the user back to the app, at long last.
+    # Just in case they've tried to be super clever and host multiple apps with the same
+    # callback URL, we'll provide the client ID along with the state
+    return redirect(app.callback_url + "?client_id=" + app.client_id + "&state=" + state)
+
+def userdata(request):
```
wilhelmklopp: This view should maybe go in a different app?
It's not specifically part of OAuth and almost more like a service that uses OAuth

frontend/gulp/tasks/fileModules.json
```diff
@@ -1,4 +1,5 @@
 {
   "dashboard": "dashboard" , 
-  "agreement": "dashboard"
+  "agreement": "dashboard",
+  "authorise": "oauth"
```
wilhelmklopp: this should probably have the same name as well?

frontend/src/react/components/appList.jsx
```diff
@@ -187,20 +208,197 @@ class App extends React.Component {
...
+    });
+  }
+
+  setSaveCallbackUrl() {
+    this.setState({
+      callbackUrlSaved: false
+    })
+  }
+
+  updateRoomBookingsScope(e){
```
wilhelmklopp: as mentioned before, we can probably be a lot more DRY by merging these separate endpoints into one.

frontend/src/react/components/appPermissions.jsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
```
wilhelmklopp: Text in here needs to be less verbose imo

frontend/src/sass/includes/rc-collapse.scss
```diff
@@ -0,0 +1,61 @@
+.rc-collapse {
+  background-color: #f7f7f7;
+  border-radius: 3px;
+  border: 1px solid #d9d9d9;
+}
+.rc-collapse-anim-active {
+  transition: height 0.2s ease-out;
+}
+.rc-collapse > .rc-collapse-item {
```
wilhelmklopp: You can be more DRY by using the nesting syntax provided by SCSS. 

That goes for all the other selectors in this file as well.

frontend/src/sass/pages/authorise.scss
```diff
@@ -0,0 +1,20 @@
...
+
+h1, h2, h3, h4, h5{
+	font-family: 'Roboto', sans-serif;
+}
+
+.centered {
+  text-align: center;
+}
+
+body{
```
wilhelmklopp: This and other global selectors should be defined already, right?

backend/uclapi/.env.example
ChristopherHammond13: Yes, it should block all ports at those domains. The production and staging blocks will also include staging.ninja and uclapi.com to prevent anybody trying to make our code redirect round in a loop.

backend/uclapi/dashboard/migrations/0006_auto_20170516_2014.py
ChristopherHammond13: I may need to do a merge migration to get in line with master, but I don't want to do that until the last moment if possible.

backend/uclapi/dashboard/models.py
ChristopherHammond13: Sure, does this explain your choice? https://stackoverflow.com/a/10052288/5297057

backend/uclapi/dashboard/urls.py
ChristopherHammond13: which is preferred for URLs?

backend/uclapi/dashboard/urls.py
ChristopherHammond13: There are some issues with that (mainly that each scope is sort of hard coded even in the model) and it would require a massive if..elif...etc. block. We may need to discuss how to scope more intelligently.

backend/uclapi/dashboard/views.py
ChristopherHammond13: Yes

backend/uclapi/oauth/models.py
ChristopherHammond13: Oh, good to know! I like having the verbosity of it though; I feel more comfortable actually seeing my PK :P

backend/uclapi/oauth/templates/permissions.html
ChristopherHammond13: Good shout

backend/uclapi/oauth/views.py
ChristopherHammond13: How do you know which are spat out? Do you look up the docs for each one?

backend/uclapi/oauth/views.py
ChristopherHammond13: is this a spec thing or a "I like it because that's how Slack does it" thing?

backend/uclapi/oauth/views.py
ChristopherHammond13: okay. That's something to implement later I think :P Going to have to revamp the client library lol

Agreed though, as it needs to spec comply

backend/uclapi/oauth/views.py
ChristopherHammond13: Agreed

backend/uclapi/oauth/views.py
ChristopherHammond13: Not sure what the best way to deal with situations like this is. Probably best to ask @Faiz

frontend/gulp/tasks/fileModules.json
ChristopherHammond13: we either need to change the template name, or this and the link to the js file in the template html file.

frontend/src/react/components/appList.jsx
ChristopherHammond13: Agreed, but again this comes down to how we intend to scale out scoping.

frontend/src/react/components/appPermissions.jsx
ChristopherHammond13: ahh, feel free to suggest better text.

frontend/src/sass/includes/rc-collapse.scss
ChristopherHammond13: I didn't write this code (it came from the rc-collapse library)

frontend/src/sass/pages/authorise.scss
ChristopherHammond13: It was copied from the dashboard scss I think, we authorise does not inherit from.

backend/uclapi/dashboard/urls.py
wilhelmklopp: I don't mind as long as we're consistent.

ChristopherHammond13: @Faiz what is the most pythonic answer to this?

backend/uclapi/oauth/models.py
wilhelmklopp: It's just a consistency thing I guess. Also remember the bulk_create problem we had a while back? that was resolved by making this field implicit.

backend/uclapi/oauth/views.py
wilhelmklopp: you should encounter each specific exception during testing. Here you're probably looking for `KeyError`

backend/uclapi/oauth/views.py
wilhelmklopp: Slack does something like this, but what I mean here is a "we should have one callback, not 4" thing

backend/uclapi/oauth/models.py
ChristopherHammond13: Interestingly, you can solve most issues with id fields (from the looks of things) by setting the values to `None`; Django will then assign an ID for you. That's probably how we could have solved that issue back then.

backend/uclapi/oauth/views.py
ChristopherHammond13: Is there a list of exceptions to catch for each function? I'm crappy at Python as I'm sure you can now see!

backend/uclapi/dashboard/urls.py
Faiz: `snake_case` is the pythonic way

backend/uclapi/oauth/views.py
ChristopherHammond13: Sounds sensible.

wilhelmklopp: image

This is from the simplified OAuth thing 😛

ChristopherHammond13: yep, yep.

backend/uclapi/oauth/app_helpers.py
```diff
@@ -0,0 +1,22 @@
+from binascii import hexlify
+import os
+
+
+def generate_user_token():
```
jermenkoo: Here you are skipping 4 letters in your key (so you use only 56 out of 60).

While I think it does not matter, the reason is quite subtle and might not be visible to the user directly.

One way how to write this function would be:
```
import os 
import textwrap


def generate_user_token():
    key = hexlify(os.urandom(30)).decode()
    dashed = '-'.join(textwrap.wrap(key, 15))
    final ="uclapi-user-" + dashed

    return final
```

Then you can observe that you have 60 hex characters instead of 56.

jermenkoo:
btw `master` is ahead again, it would be good to merge it back in

backend/uclapi/oauth/views.py
```diff
@@ -0,0 +1,403 @@
...
+    try:
+        data = json.loads(raw_data_str)
+    except ValueError:
+        return JsonResponse({
+            "ok": False,
+            "error": ("The JSON data was not in the expected format."
+                      " Please contact support.")
+        })
+
+    user = User.objects.get(employee_id=data["user_upi"])
```
jermenkoo: @ChristopherHammond13 also I had no clue what to do with this

ChristopherHammond13:
I am currently reimplementing how the scoping system works and refactoring a bunch of it whilst I'm going so this won't be merged yet.

backend/uclapi/oauth/scoping.py
```diff
@@ -0,0 +1,69 @@
...
+
+        return ((1 << scope_shift) & current) > 0
+
+    # Remove a scope from the current scope number
+    def remove_scope(self, current, scope_name):
+        try:
+            scope_shift = self.SCOPE_MAP[scope_name][0]
+        except KeyError:
+            return current
+
```
jermenkoo: The lines below might be better written as `current & ~(1 << scope_shift)`. If you try to unset a bit which is NOT set, nothing will happen anyways.

backend/uclapi/oauth/views.py
```diff
@@ -0,0 +1,402 @@
...
+            full_name=display_name,
+            given_name=given_name,
+            department=department,
+            cn=cn,
+            raw_intranet_groups=groups,
+            employee_id=employee_id
+        )
+
+        user.save()
+        keen.add_event("signup", {
```
jermenkoo: This will fail as `keen` is not imported. (removed in `c7d49644e90cbd7333d0601fa37cbfbe05007af0`)

backend/uclapi/oauth/views.py
```diff
@@ -0,0 +1,402 @@
...
+        return JsonResponse({
+            "ok": False,
+            "error": ("The JSON data was not in the expected format."
+                      " Please contact support.")
+        })
+
+    app = App.objects.get(client_id=data["client_id"])
+    state = data["state"]
+
+    redir = app.callback_url + "?result=denied?state=" + state
```
jermenkoo: should the second `?` be a `&`?

ChristopherHammond13:
I have some unstaged react changes locally so that the dashboard reflects these new changes then we should be able to merge :) +1

ChristopherHammond13:
If anybody wants to finish this, please do!
Critical lines:
- https://github.com/uclapi/uclapi/blob/OAuth/frontend/src/react/components/appList.jsx#L37
- https://github.com/uclapi/uclapi/blob/OAuth/frontend/src/react/components/appList.jsx#L311
- https://github.com/uclapi/uclapi/blob/OAuth/frontend/src/react/components/appList.jsx#L571

backend/uclapi/oauth/decorators.py
```diff
@@ -0,0 +1,71 @@
...
+            hmac_b64 = base64.b64encode(hmac_digest).decode()
+            if client_secret_proof != hmac_b64:
+                response = JsonResponse({
+                    "ok": False,
+                    "error": "Client secret HMAC verification failed."
+                })
+                response.status_code = 400
+                return response
+
+            s = Scopes()
```
jermenkoo: This gets shadowed away by the next line?

backend/uclapi/oauth/scoping.py
```diff
@@ -0,0 +1,111 @@
+# Storage of the scope map
+# The purpose of this setup is that the OAuth scope of any app can be stored in a single field,
+# and it can be added to in the future.
+# We have a BigIntegerField to work with, which means 64 bits of storage. This translates into
+# 64 types of scope, each of which can be checked with a bit mask.
+# E.g. roombookings has scope 0, which is 0000000000000000000000000000000000000000000000000000000000000001b.
+# This is because the 0th bit (LSB) is set to 1.
+# Room bookings + UCLU = 1001b, or a scope number of 11
```
jermenkoo: Roombookings + UCL = 0b101 which is 5

backend/uclapi/dashboard/api_applications.py
```diff
@@ -182,3 +197,187 @@ def delete_app(request):
...
+        keen.add_event("App scopes changed", {
+            "appid": app_id,
+            "userid": user.id,
+            "scopes": scopes
+        })
+
+        return PrettyJsonResponse({
+            "success": True,
+            "message": "Scope successfully changed",
+        })
```
ChristopherHammond13: this is a lie

backend/uclapi/oauth/tests.py
```diff
@@ -0,0 +1,63 @@
...
+        self.assertEqual(scopes_dict, [
+            {
+                "name": "roombookings",
+                "id": 0
+            },
+            {
+                "name": "uclu",
+                "id": 2
+            }
+        ])
```
ChristopherHammond13: this is also a lie

backend/uclapi/oauth/tests.py
```diff
@@ -0,0 +1,191 @@
...
+        )
+        app_ = App.objects.create(user=user_, name="An App")
+        oauth_scope = OAuthScope.objects.create()
+        oauth_token = OAuthToken.objects.create(
+            app=app_, user=user_, scope=oauth_scope
+        )
+
+        # TODO: THIS WILL GET CHANGED
+        # calculate the client_secret_proof
+        import hmac, hashlib, base64
```
jermenkoo: @ChristopherHammond13 I had no idea how to get / obtain / calculate `client_secret_proof` so I calculated it like this - yes or no?

backend/uclapi/dashboard/urls.py
wilhelmklopp: this still needs to be fixed it seems

backend/uclapi/dashboard/urls.py
```diff
@@ -11,5 +12,7 @@
     url(r'api/rename/$', rename_app),
     url(r'api/regen/$', regenerate_app_token),
     url(r'api/delete/$', delete_app),
+    url(r'api/setcallbackurl/$', set_callback_url),
+    url(r'api/updatescopes/$', update_scopes),
```
wilhelmklopp: and the same consistency problem here

backend/uclapi/dashboard/views.py
```diff
@@ -71,7 +72,8 @@ def dashboard(request):
         user_id = request.session["user_id"]
     except KeyError:
         url = os.environ["SHIBBOLETH_ROOT"] + "/Login?target="
-        param = request.build_absolute_uri(request.path) + "user/login.callback"
+        param = (request.build_absolute_uri(request.path) +
+                 "user/login.callback")
```
wilhelmklopp: I think we should decide what URL naming we should use going forward.

The dot syntax is the third different style out of the ones we've used so far haha

backend/uclapi/oauth/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
```
wilhelmklopp: It looks like we don't need this file. Can we delete it?

backend/uclapi/oauth/templates/permissions.html
wilhelmklopp: This also still needs to be adressed?

frontend/src/react/components/appPermissions.jsx
```diff
@@ -0,0 +1,49 @@
...
+                <li>Your Email Address ({this.props.user_email})</li>
+                <li>Your Department ({this.props.user_department})</li>
+                <li>Your UPI ({this.props.user_upi})</li>
+            </ul>
+
+            <h4>Personal Data Permissions</h4>
+             {
+                (this.props.scopes.length == 0) &&
+                <em>
+                    This app has not requested access to any of your personal UCL data. It will only be able to see public data, such as the public timetable and room bookings information. If you are expecting this app to show you any personal data, please contact the vendor.
```
wilhelmklopp: Strike this `If you are expecting this app to show you any personal data, please contact the vendor.` imo

Adds too much information at this stage

frontend/src/react/pages/authorise.jsx
```diff
@@ -0,0 +1,57 @@
...
+            </form>
+            
+            <form method="post" action="/oauth/user/deny" style={{display: 'inline'}}>
+              <button className="pure-button button-error padded">Deny</button>
+              <input type="hidden" name="signed_app_data" value={this.state.data.signed_data}/>
+              <input type="hidden" name="csrfmiddlewaretoken" value={Cookies.get('csrftoken')}/>
+            </form>
+            <hr/>
+            <em>
+              Apps you connect to UCL via the API (<a href="https://uclapi.com/">uclapi.com</a>) cannot access any data that is not shown above. Your personal details such as your home address, date of birth, UCL password and phone number are kept private and cannot be accessed even with these permissions. If you have any further questions please do not hesitate to contact us by email: <a href="mailto:isd.apiteam@ucl.ac.uk">isd.apiteam@ucl.ac.uk</a>.
```
wilhelmklopp: Change this to:
```
{`${this.state.data.app_name} can only access the data shown above. Other personal details that UCL stores for you are kept private and secure. Have a question? Contact us at <a href="mailto:isd.apiteam@ucl.ac.uk">isd.apiteam@ucl.ac.uk</a>.`}
```

frontend/src/sass/includes/rc-collapse.scss
wilhelmklopp: If this is a library, can we not import it as a react/gulp module? Seems wrong to track it in git

frontend/src/sass/pages/authorise.scss
wilhelmklopp: In that case we should make a global scss file and have dashboard and authorise inherit from it right?

backend/uclapi/dashboard/views.py
ChristopherHammond13: I agree

backend/uclapi/oauth/models.py
```diff
@@ -0,0 +1,46 @@
...
+
+    # The scope that the key works within. Every key has a unique scope
+    # (in case the developer requests more permissions later)
+    scope = models.OneToOneField(
+        OAuthScope,
+        on_delete=models.CASCADE
+    )
+
+    # Whether the token is active or not
+    active = models.BooleanField(default=True)
```
ChristopherHammond13: fixed

backend/uclapi/oauth/views.py
```diff
@@ -0,0 +1,412 @@
...
+            full_name=display_name,
+            given_name=given_name,
+            department=department,
+            cn=cn,
+            raw_intranet_groups=groups,
+            employee_id=employee_id
+        )
+
+        user.save()
+        keen_add_event.delay("signup", {
```
ChristopherHammond13: This should be fixed now.

ChristopherHammond13:
This is now in staging. We need to confirm prior to merging into master which services we'll have available so that I can set the `scopes` list to something initially. Each new integration after this will need to provide a scope description for the dashboard and OAuth backend.

PR #90, opened 2017-06-13
Anirudh/get started page

anirudhpillai:
## What does this PR do?
New homepage for uclapi.com.
Also, new Django model for temporary token.
The temp token is used on the client side and has limited capabilities (https://trello.com/c/qWUOjK5t/248-public-api-demo).

## ✅ Pull Request checklist

- [x] Is this code complete?

No

## :squirrel: Deploy notes
Need to run migrations as part of deployment as it has code for the temporary token.


backend/uclapi/roombookings/decorators.py
```diff
@@ -1,15 +1,81 @@
...
+            except ObjectDoesNotExist:
+                response = JsonResponse({
+                    "ok": False,
+                    "error": "Invalid temporary token"
+                })
+                response.status_code = 400
+                return response
+
+            if request.path != "/roombookings/bookings":
+                temp_token.uses += 1
```
wilhelmklopp: should we increment the uses even if there is no "use" because we returned an error?
Just asking, this might totally be the right thing to do :)

backend/uclapi/roombookings/decorators.py
```diff
@@ -51,6 +114,33 @@ def wrapped(request, *args, **kwargs):
...
         queryparams = dict(request.GET)
 
+        token = request.GET["token"]
+
+        is_temp_token = None
+
+        try:
+            if token.split("-")[1] == "temp":
+                is_temp_token = True
+        except:
```
wilhelmklopp: same here. Explicit except would be great

backend/uclapi/roombookings/decorators.py
```diff
@@ -64,11 +154,25 @@ def wrapped(request, *args, **kwargs):
...
 def throttle(view_func):
     def wrapped(request, *args, **kwargs):
         token = request.GET.get("token")
+
+        is_temp_token = None
+
+        try:
+            if token.split("-")[1] == "temp":
+                is_temp_token = True
+        except:
```
wilhelmklopp: and here as well :P

backend/uclapi/dashboard/models.py
```diff
@@ -36,7 +38,7 @@ class App(models.Model):
     last_updated = models.DateTimeField(auto_now=True)
 
     def regenerate_token(self):
-        new_token = generate_api_token()
+        new_token = generate_temp_api_token()
```
anirudhpillai: You're right it shouldn't

backend/uclapi/dashboard/views.py
```diff
@@ -122,3 +122,21 @@ def dashboard(request):
...
+    try:
+        user_id = request.session["user_id"]
+    except KeyError:
+        logged_in = False
+
+    temp_token = TemporaryToken.objects.create()
+    return render(request, 'getStarted.html', {
+        'initial_data': {
+            'temp_token': temp_token.api_token,
+            'logged_in': str(logged_in)
```
anirudhpillai: yes, its because of how Django templates work. I couldn't pass a boolean

backend/uclapi/roombookings/decorators.py
anirudhpillai: I thought we should because we won't be using the temp token for some other route, it would only be people trying to exploit it

backend/uclapi/dashboard/app_helpers.py
```diff
@@ -2,7 +2,7 @@
 import os
 
 
-def generate_api_token():
+def generate_api_token(temp=False):
```
anirudhpillai: True, don't know how I missed that

backend/uclapi/roombookings/decorators.py
wilhelmklopp: OK! :)

backend/uclapi/dashboard/models.py
```diff
@@ -45,6 +47,24 @@ class Meta:
         _DATABASE = 'default'
 
 
+class TemporaryToken(models.Model):
+    id = models.CharField(
```
anirudhpillai: @wilhelmklopp this is the only thing that's left to be fixed. 
Django wasn't allowing me to run the migration after I updated it.

backend/uclapi/roombookings/decorators.py
```diff
@@ -24,6 +26,66 @@ def wrapped(request, *args, **kwargs):
             return response
 
         try:
+            if token.split("-")[1] == "temp":
+                is_temp_token = True
+        except IndexError:
+            is_temp_token = False
+
+        if is_temp_token:
```
Faiz: this could throw an undefined error, because in one case the `is_temp_token` can go undefined

backend/uclapi/roombookings/decorators.py
```diff
@@ -24,6 +26,66 @@ def wrapped(request, *args, **kwargs):
...
+                temp_token.delete()  # Delete expired token
+                response = PrettyJsonResponse({
+                    "ok": False,
+                    "error": "Temporary token expired"
+                })
+                response.status_code = 400
+                return response
+
+            if not request.GET._mutable:
+                request.GET._mutable = True
```
Faiz: cant you just set it to `True` haha

backend/uclapi/roombookings/decorators.py
```diff
@@ -55,6 +114,31 @@ def wrapped(request, *args, **kwargs):
...
+
+        try:
+            if token.split("-")[1] == "temp":
+                is_temp_token = True
+        except IndexError:
+            is_temp_token = False
+
+        user = None
+
+        if is_temp_token:
```
Faiz: same undefined error

backend/uclapi/roombookings/decorators.py
```diff
@@ -68,12 +152,23 @@ def wrapped(request, *args, **kwargs):
...
     def wrapped(request, *args, **kwargs):
         token = request.GET.get("token")
+
+        try:
+            if token.split("-")[1] == "temp":
+                is_temp_token = True
+        except IndexError:
+            is_temp_token = False
+
+        if is_temp_token:
```
Faiz: again

backend/uclapi/roombookings/decorators.py
jermenkoo: it will throw it anyways, since it is in a different scope (!)

backend/uclapi/roombookings/decorators.py
jermenkoo: same as above

backend/uclapi/roombookings/decorators.py
jermenkoo: same as above

backend/uclapi/roombookings/decorators.py
Faiz: Try except scope is one level up. But that's not the case for if statement

backend/uclapi/roombookings/decorators.py
jermenkoo: what does this do anyways? @anirudhpillai

backend/uclapi/roombookings/decorators.py
```diff
@@ -23,6 +25,67 @@ def wrapped(request, *args, **kwargs):
...
+                    "ok": False,
+                    "error": "Temporary token expired"
+                })
+                response.status_code = 400
+                return response
+
+            if not request.GET._mutable:
+                request.GET._mutable = True
+
+            request.GET['results_per_page'] = 1
```
jermenkoo: shouldn't this go somewhere else? sep. of concerns

backend/uclapi/roombookings/decorators.py
anirudhpillai: That is required so that I can modify the request variable like in `request.GET['results_per_page'] = 1`

backend/uclapi/roombookings/decorators.py
anirudhpillai: Thanks for fixing it Jaro

backend/uclapi/roombookings/decorators.py
anirudhpillai: Not sure, what we want is that temp token should only return one listing so as to prevent abuse. I thought it would be useful to put all the temp token code in here.

backend/uclapi/roombookings/decorators.py
Faiz: yeah but just do `request.GET._mutable = True` you dont need an if statement there

anirudhpillai: haha, sure I'll change that, had just copied it from stackoverflow.

backend/uclapi/roombookings/decorators.py
jermenkoo: alright, fine with me then!

backend/uclapi/roombookings/decorators.py
jermenkoo: L125-L126 might never get called since you check for that already in `does_token_exist` which means at this point you already checked if it exists & possibly throttled it so here you either have a valid {temp, permanent} token

backend/uclapi/dashboard/views.py
```diff
@@ -124,3 +124,21 @@ def dashboard(request):
...
         'initial_data': initial_data
     })
+
+
+@ensure_csrf_cookie
+def get_started(request):
+    logged_in = True
+
+    try:
+        user_id = request.session["user_id"]
```
jermenkoo: this is never used anywhere @anirudhpillai

jermenkoo:
Maybe `generate_temp_api_token` could be changed like this:

```
import os 
import textwrap


def generate_user_token():
    key = hexlify(os.urandom(30)).decode()
    dashed = '-'.join(textwrap.wrap(key, 15))
    final = "uclapi-user-" + dashed

    return final
```

Chris implemented the same thing here: https://github.com/uclapi/uclapi/pull/84/commits/b456e982e1cbc902fa1aefaf221b058edb6c778f

Later we could extract them and put it into one file.

ChristopherHammond13:
Going to squash and merge this into master now. Thanks for everything, @anirudhpillai! Toooo stagingggg!!!!

PR #91, opened 2017-06-21
Webhooks

wilhelmklopp:
## What does this PR do?
This PR adds webhooks, so us (UCL API) making HTTP requests to URLs that integrators register.

## ✅ Pull Request checklist

- [x] Is this code complete?

https://github.com/uclapi/apiDocs/pull/22

No

## :squirrel: Deploy notes
- Will need to run migrations

## Anything else
Todo prior to webhooks public launch:
- [ ] Clean up webhook_demo_app & add readme
- [ ] Write Getting Started medium post to be published on launch


jermenkoo:
`requirements.txt` needs to be updated & there's a conflicting migration right now - will review code tonight

ChristopherHammond13:
@jermenkoo the requirements in his branch look up to date. We should solve migration conflicts once we merge OAuth into master to avoid working twice.

jermenkoo:
@ChristopherHammond13 https://github.com/uclapi/uclapi/pull/91/files#diff-810343d063ddfe8e844ae77d2f8ff5ef everything you updated gets downgraded to older version?

hennersz:
Frontend looks good to me

backend/uclapi/dashboard/webhook_views.py
```diff
@@ -0,0 +1,203 @@
...
+        if "challenge" not in resp.keys():
+            return False
+        return (
+            resp["challenge"] ==
+            ownership_challenge
+        )
+
+
+def is_url_safe(url):
+    if url[:8] != "https://":
```
jermenkoo: `if not url.startswith("https://"):`

backend/uclapi/dashboard/tests.py
```diff
@@ -1,9 +1,18 @@
+import json
 from unittest.mock import patch
 
 from django.test import RequestFactory, TestCase
+from rest_framework.test import APIRequestFactory
 
 from .middleware.fake_shibboleth_middleware import FakeShibbolethMiddleWare
 from .models import App, User
+from .webhook_views import (
```
jermenkoo: alphabetical order pref.

backend/uclapi/roombookings/management/commands/trigger_webhooks.py
```diff
@@ -0,0 +1,123 @@
...
+        new_bookings = _serialize_bookings(
+            new_booking_table.objects.filter(
+                startdatetime__gt=now
+            )
+        )
+
+        ddiff = DeepDiff(old_bookings, new_bookings, ignore_order=True)
+
+        webhooks = Webhook.objects.all()
+        #  assumption: list of webhooks will be longer than ddiff
```
jermenkoo: What happens if it will NOT be? Is it somewhere documented?

wilhelmklopp: It'll still work, just take longer.

jermenkoo: 👍

wilhelmklopp:
Webhooks is ready to go live 😎 

@ChristopherHammond13 when we have one approving code review, could you merge this?

backend/uclapi/dashboard/models.py
```diff
@@ -96,3 +101,43 @@ class APICall(models.Model):
 
     class Meta:
         _DATABASE = 'default'
+
+
+class Webhook(models.Model):
```
ChristopherHammond13: Should we be placing Webhook logic in the dashboard models file? I feel like it's the same thing that you called me out on :P

backend/uclapi/dashboard/webhook_views.py
```diff
@@ -0,0 +1,187 @@
...
+        webhook.url = url
+        webhook.save()
+
+    webhook.siteid = siteid
+    webhook.roomid = roomid
+    webhook.contact = contact
+    webhook.enabled = True
+    webhook.save()
+
+    keen.add_event("Webhook edited", {
```
ChristopherHammond13: This should use the delay task to run via Celery now that we have that configured

backend/uclapi/dashboard/webhook_views.py
```diff
@@ -0,0 +1,187 @@
+from django.http import JsonResponse
+from .models import App, User
+import keen
```
ChristopherHammond13: Similar to the comment below, we never use keen directly now to avoid requests taking too long

backend/uclapi/dashboard/webhook_views.py
```diff
@@ -0,0 +1,187 @@
...
+from django.core.exceptions import ObjectDoesNotExist
+import requests
+from .app_helpers import generate_secret, is_url_safe
+
+
+def user_owns_app(user_id, app_id):
+    user = User.objects.get(id=user_id)
+    try:
+        app = App.objects.get(id=app_id)
+    except ObjectDoesNotExist:
```
ChristopherHammond13: Prefer `App.DoesNotExist` here for clarity, but I'm just being picky lol

backend/uclapi/dashboard/webhook_views.py
```diff
@@ -0,0 +1,187 @@
+from django.http import JsonResponse
```
ChristopherHammond13: Prefer to use PrettyJsonResponse here. You can import Jaro's prettifier as JsonResponse to avoid rewriting code below. It's easier for debugging / viewing in a browser 👍

frontend/src/react/components/app/defaultHeaders.js
```diff
@@ -0,0 +1,6 @@
+import Cookies from 'js-cookie';
```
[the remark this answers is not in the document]
ChristopherHammond13: I'm not sure how to fix these notices. @jermenkoo ?

frontend/yarn.lock
```diff
@@ -0,0 +1,4240 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
ChristopherHammond13: I know nothing about yarn. Do we need this file to build the frontend now? I've been using `gulp`, but has this now changed?

ChristopherHammond13:
The most important thing to confirm here is that checkboxes for scopes will still work on the frontend (as `master` currently has an empty OAuth `SCOPE_MAP` dictionary, so no scope checkboxes are rendered on the frontend).

When I originally wrote my horrendous React code I had to hack around the checkboxes because selecting them triggered a complete re-render of the checkboxes which reset their check state to the one defined in HTML. If this works, and we clean up what I recommended above, I'm happy to get this into staging.

frontend/src/react/components/app/defaultHeaders.js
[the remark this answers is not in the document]
wilhelmklopp: yeah this is a hound specific error. All our stuff goes through babel

wilhelmklopp:
@ChristopherHammond13 did you change the OAuth code to account for `deleted=True` of App?
That change might look like something like `App.objects.filter(user=user, deleted=False)`

backend/uclapi/dashboard/models.py
wilhelmklopp: 😜 
There's no separate app for webhooks (it's part of `dashboard`) and I think what I mentioned in the OAuth PR is oauth specific fields being part of the same table as `app`, but in this case webhook specific data is in its own table.

backend/uclapi/dashboard/models.py
```diff
@@ -70,3 +75,43 @@ class APICall(models.Model):
...
+    enabled = models.BooleanField(default=False)
+    last_updated = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        _DATABASE = 'default'
+
+
+class WebhookTriggerHistory(models.Model):
+    webhook = models.ForeignKey(Webhook)
+    payload = models.CharField(max_length=10000000)
```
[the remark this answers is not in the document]
wilhelmklopp: hmm I think some length constraint _is_ useful here. It stops weird cases where we get like 15GB stored in a single field.

backend/uclapi/dashboard/webhook_views.py
wilhelmklopp: Agreed, that's much more beautiful and allows us to remove one line of code 💪

frontend/yarn.lock
wilhelmklopp: Nope, nothing we need to change. yarn is just an alternative to npm I have been using.
https://yarnpkg.com

wilhelmklopp:
> The most important thing to confirm here is that checkboxes for scopes will still work on the frontend (as master currently has an empty OAuth SCOPE_MAP dictionary, so no scope checkboxes are rendered on the frontend).

> When I originally wrote my horrendous React code I had to hack around the checkboxes because selecting them triggered a complete re-render of the checkboxes which reset their check state to the one defined in HTML. If this works, and we clean up what I recommended above, I'm happy to get this into staging.

If there is anything broken here (and that means it'd already be broken on master) I'd prefer to fix that up in a PR that actually introduces a scope and thus where this functionality comes into play. The `timetable` PR might be a good contendor for that.

wilhelmklopp:
Thanks for the CR @ChristopherHammond13 
From what I can tell this is ready to go now

backend/uclapi/dashboard/models.py
ChristopherHammond13: No problem then  ;P

backend/uclapi/dashboard/models.py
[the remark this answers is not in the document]
ChristopherHammond13: is that ever a possibility?

frontend/yarn.lock
ChristopherHammond13: fab :)

backend/uclapi/dashboard/models.py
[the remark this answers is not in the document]
wilhelmklopp: I mean in principle it's a failsafe, especially given that we don't control the data source (Estates/ISD does). We should not topple even if something gets messed up upstream.

Additionally, if the buckets aren't set up properly (if for example only one is set up), then running `update_gencache` would trigger a webhook event with all roombookings in history (which should obviously fail). The max_length on this field makes it fail, so that's what we want 👍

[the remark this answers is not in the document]
ChristopherHammond13: Sounds good then :)

ChristopherHammond13:
Let's hit staging!

PR #94, opened 2017-07-04
Load homepage pictures from our own server

wilhelmklopp:
## What does this PR do?
Currently we're loading these images on the homepage, and we're getting mixed security warnings because of that. This PR fixes that.

## ✅ Pull Request checklist

- [X] Is this code complete?
- [ ] ~Are tests done/modified?~

not applicable

No

## Anything else
If we change our django `STATIC_URL`, this will break. There might be a way to account for this in a better way, but I couldn't think of anything off the top of my head.

## :squirrel: Deploy notes
Shouldn't require anything special

This is ready to go to staging whenever @ChristopherHammond13 :)


wilhelmklopp:
@ChristopherHammond13 
I might be doing something wrong with the getStarted.js build file? The number of additions in the diff seems wrong: https://github.com/uclapi/uclapi/pull/94/files#diff-89eac3821f1510cf5ff2c199f9237d7d

ChristopherHammond13:
@wilhelmklopp not sure why you removed the statics gitignore. We do a `collectstatic` on the server so they're ok to merge; we just need the result of `NODE_ENV=production gulp productionBuild` to be committed to avoid having to run gulp, node, etc on our web servers. 

Going to have dinner then take a closer look at this. It might be better to use a folder name like `images` which won't cause issues?

wilhelmklopp:
Removed the `/static/` because I couldn't commit these files otherwise.

And the images are in a folder called images. Django serves everything that `collectstatic` collects from a STATIC_URL, which is defined in settings.py

ChristopherHammond13:
@wilhelmklopp you can customise multiple static URLs. Perhaps we should add another for stuff we need to commit? 
https://docs.djangoproject.com/en/dev/ref/settings/#std:setting-STATICFILES_DIRS

ChristopherHammond13:
This should sort the issues out now that I've moved those images out of the `statics` folder and into the images one

PR #96, opened 2017-07-06
Migrate Keen Reporting to Celery

ChristopherHammond13:
## What does this PR do?
Adds support for a Redis-backed Celery setup to use with the API. This allows long-running background tasks, such as the Keen reporting, to be done by a Celery worker with a shared Redis cluster.

This should shave nearly a second off some requests which are being held up currently by a blocking call to Keen.

This also updates a bunch of our requirements including Django, Requests and some of the other libraries.

## ✅ Pull Request checklist

- [x] Is this code complete?

No
Unless they have built their clients to expect a long wait time for results, in which case they'll be out of luck ;)

## :squirrel: Deploy notes
- `pip install -r requirements.txt`
- Needs a new Supervisor config file to make the Celery worker actually load at boot
- `supervisorctl restart uclapi` once the Celery worker is running

## Anything else
Expect **SPEED**

No tests written because there's no business logic here, just implementation.


.travis.yml
```diff
@@ -24,6 +24,8 @@ before_script:
   # Install cx_Oracle (which depends on Oracle IC)
   - cd ./../backend/uclapi
   - pip3 install $(cat requirements.txt | grep "cx-Oracle")
+  - sudo apt-get -y install redis-server
+  - sudo service redis-server restart
```
wilhelmklopp: Do we need these two lines? We already have redis set up as a `service` on line 7
Docs for that: https://docs.travis-ci.com/user/database-setup/#Redis

backend/uclapi/dashboard/tasks.py
```diff
@@ -0,0 +1,20 @@
...
+@shared_task
+def test_task(param):
+    return 'The test task executed with argument "%s" ' % param
+
+
+@shared_task
+def keen_add_event_task(title, data):
+    try:
+        keen.add_event(title, data)
+    except Exception:
```
wilhelmklopp: I feel like this exception should be removed. If keen is down or there is a different reason something is failing, we should be seeing that as an exception in Sentry, so that we can action it.

backend/uclapi/roombookings/decorators.py
```diff
@@ -152,7 +152,8 @@ def wrapped(request, *args, **kwargs):
             "queryparams": queryparams
         }
 
-        keen.add_event("apicall", parameters)
+        print("Logging apicall")
```
wilhelmklopp: This print should be removed

backend/uclapi/uclapi/settings.py
```diff
@@ -198,3 +198,9 @@
     FAIR_USE_POLICY = list(fp)
 
 REDIS_UCLAPI_HOST = os.environ.get("REDIS_UCLAPI_HOST")
+
+# Celery Settings
+CELERY_BROKER_URL = 'redis://' + REDIS_UCLAPI_HOST + ':6379/0'
```
wilhelmklopp: could the environment variable here contain both the `redis://` and the `:6379/0`? That seems like it will scale better

I wonder if we can give the environment variable itself a better name too? Maybe something like `CELERY_BROKER_URL`?
I think we also need an update to the .env.example

.travis.yml
jermenkoo: No

backend/uclapi/uclapi/settings.py
ChristopherHammond13: We have to have them separate because `StrictRedis` takes just the server as a parameter, so we would have to store the same value twice.

backend/uclapi/dashboard/tasks.py
ChristopherHammond13: I haven't decided on this just yet. Are there guidelines on this?

backend/uclapi/uclapi/settings.py
jermenkoo: > all fields after the scheme are optional, and will default to localhost on port 6379, using database 0.
You can omit the other things @ChristopherHammond13, they default to 6379 and 0 for throttling anyways.

ChristopherHammond13: made this change now and it seems ok on dev :)

ChristopherHammond13:
Made a bunch of changes to this now. @wilhelmklopp or @jermenkoo could one of you please review? :)

backend/uclapi/dashboard/tasks.py
wilhelmklopp: We still need to fix this. Just removing the try except block should do it.

backend/uclapi/uclapi/celery.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+read_dotenv(os.path.join(os.path.dirname(os.path.dirname(__file__)), '.env'))
+
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'uclapi.settings')
+
+
+class Celery(celery.Celery):
+    def on_configure(self):
+        if os.environ.get("SENTRY_DSN") != "":
```
wilhelmklopp: If the environment variable here is not set, that might show a weird error.
Probably best to use `os.environ["SENTRY_DSN"]` ?

backend/uclapi/uclapi/settings.py
```diff
@@ -198,3 +198,9 @@
     FAIR_USE_POLICY = list(fp)
 
 REDIS_UCLAPI_HOST = os.environ.get("REDIS_UCLAPI_HOST")
+
+# Celery Settings
+CELERY_BROKER_URL = 'redis://' + REDIS_UCLAPI_HOST
+CELERY_ACCEPT_CONTENT = ['json']
+CELERY_TASK_SERIALIZER = 'json'
+CELERY_RESULT_SERIALIZER = 'json'
```
wilhelmklopp: Do we need this line?
Just wondering :)

ChristopherHammond13: the setup tutorials suggest it so I'm just gonna roll with it

backend/uclapi/uclapi/celery.py
ChristopherHammond13: easy to fix, will do once I've sorted OAuth nonces out

backend/uclapi/dashboard/tasks.py
Faiz: explicit errors should be here. @wilhelmklopp has already mentioned it. There should be a possible list of exceptions in keep docs.

backend/uclapi/uclapi/celery.py
```diff
@@ -0,0 +1,53 @@
...
+    register_handlers as opbeat_register_handlers, \
+    logger as opbeat_logger  # noqa: E402#
+
+from opbeat.contrib.celery import \
+    register_signal as opbeat_register_signal  # noqa: E402
+
+
+try:
+    opbeat_register_signal(app)
+except Exception as e:
```
Faiz: hm no. check for the opbeat_register_signal possible exceptions.

ChristopherHammond13: I have resolved this now

PR #98, opened 2017-09-07
OAuth

wilhelmklopp:
## What does this PR do?
For example: Changes the way we retrieve bookings.

## ✅ Pull Request checklist

- [x] Is this code complete?

Yes/No

## :squirrel: Deploy notes


backend/uclapi/dashboard/migrations/0010_unique_ids.py
```diff
@@ -0,0 +1,35 @@
...
+            ),
+        ),
+        migrations.AlterField(
+            model_name='app',
+            name='client_secret',
+            field=models.CharField(
+                default=dashboard.app_helpers.generate_app_client_secret,
+                max_length=64,
+                unique=True
+            ),
```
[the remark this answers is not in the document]
wilhelmklopp: @jermenkoo can we turn off hound for migration files?

wilhelmklopp:
@ChristopherHammond13 was there anything we could do here to improve test coverage?

backend/uclapi/dashboard/api_applications.py
```diff
@@ -183,3 +207,152 @@ def delete_app(request):
...
+            "success": False,
+            "message": ("The requested callback URL"
+                        " does not use an acceptable protocol.")
+        })
+        response.status_code = 400
+        return response
+
+    url_data = tldextract.extract(new_callback_url)
+
+    if any(
```
wilhelmklopp: Have you seen how I did url safety verification for webhooks?

https://github.com/uclapi/uclapi/blob/f926598a10b86d0633de83959f6dcfaf1ae2a21c/backend/uclapi/dashboard/webhook_views.py#L45-L57

It seems slightly easier and more concise than the method used here

backend/uclapi/dashboard/api_applications.py
```diff
@@ -183,3 +207,152 @@ def delete_app(request):
...
+        }))
+
+    try:
+        app_id = request.POST["app_id"]
+        scopes_json = request.POST["scopes"]
+        user_id = request.session["user_id"]
+    except KeyError:
+        response = PrettyJsonResponse({
+            "success": False,
+            "message": "Request does not have app_id."
```
wilhelmklopp: message should include that there may not be scopes attached

backend/uclapi/dashboard/api_applications.py
```diff
@@ -183,3 +207,152 @@ def delete_app(request):
...
+        })
+        response.status_code = 400
+        return response
+
+    try:
+        scopes = json.loads(scopes_json)
+    except:
```
wilhelmklopp: should have explicit except

backend/uclapi/dashboard/api_applications.py
```diff
@@ -183,3 +207,152 @@ def delete_app(request):
...
+            for scope in scopes:
+                if "checked" in scope and scope["checked"]:
+                    current = s.add_scope(current, scope["name"])
+                else:
+                    current = s.remove_scope(current, scope["name"])
+
+            app.scope.scope_number = current
+            app.scope.save()
+            app.save()
+        except:
```
wilhelmklopp: should have explicit except

backend/uclapi/dashboard/app_helpers.py
```diff
@@ -30,3 +32,18 @@ def generate_app_id():
...
 
     return final
+
+
+def generate_app_client_id():
+    sr = SystemRandom()
+
+    client_id = ''.join(str(sr.randint(0, 9)) for _ in range(16))
+    client_id += "."
+    client_id += ''.join(str(sr.randint(0, 9)) for _ in range(16))
```
wilhelmklopp: could you use f-strings or .format() syntax here? It's slightly more pythonic.

backend/uclapi/dashboard/migrations/0008_populate_existing_secrets.py
```diff
@@ -0,0 +1,27 @@
...
+    for app in App.objects.all():
+        app.client_id = generate_app_client_id()
+        app.client_secret = generate_app_client_secret()
+        app.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dashboard', '0008_auto_20170701_1626'),
```
wilhelmklopp: is the fact that both this file and the previous file start with "0008" going to confuse Django? Or us for that matter?

backend/uclapi/dashboard/models.py
```diff
@@ -38,6 +40,30 @@ class App(models.Model):
     created = models.DateTimeField(auto_now=False, auto_now_add=True)
     last_updated = models.DateTimeField(auto_now=True)
 
+    client_id = models.CharField(
```
wilhelmklopp: should we create these OAuths specific fields on the App model? I think I'd prefer if they were in a separate model with a OneToOne relationship to `App`

backend/uclapi/dashboard/urls.py
```diff
@@ -11,5 +12,7 @@
     url(r'api/rename/$', rename_app),
     url(r'api/regen/$', regenerate_app_token),
     url(r'api/delete/$', delete_app),
+    url(r'api/setcallbackurl/$', set_callback_url),
+    url(r'api/updatescopes/$', update_scopes),
```
wilhelmklopp: I feel like `setcallbackurl` and `updatescopes` should be camelCased, so that they're more obvious to the user

backend/uclapi/oauth/models.py
```diff
@@ -0,0 +1,46 @@
+from django.db import models
+
+from .app_helpers import generate_user_token
+
+
+class OAuthScope(models.Model):
+    # We should really have a primary key for the OneToOneField
+    id = models.AutoField(primary_key=True)
```
wilhelmklopp: this is added by Django implicitly, just FYI 😊

backend/uclapi/oauth/models.py
```diff
@@ -0,0 +1,46 @@
...
+
+    # The scope that the key works within. Every key has a unique scope
+    # (in case the developer requests more permissions later)
+    scope = models.OneToOneField(
+        OAuthScope,
+        on_delete=models.CASCADE
+    )
+
+    # Whether the token is active or not
+    active = models.BooleanField(default=True)
```
wilhelmklopp: in what cases is a token not active? just tryna understaaand 😬

backend/uclapi/oauth/scoping.py
```diff
@@ -0,0 +1,126 @@
...
+# This is because the 0th bit (LSB) is set to 1.
+# roombookings + uclu = 101b, or a scope number of 2^2 + 2^0 = 4 + 1 = 5
+
+
+class Scopes:
+    SCOPE_MAP = {
+        "roombookings": (0, "Private room bookings data"),
+        "timetable": (1, "Private timetable data"),
+        "uclu": (2, "Private UCLU data"),
+        "moodle": (3, "Private Moodle data")
```
wilhelmklopp: I think we should remove all scopes in the scope map that aren't actually in use.

frontend/src/react/components/appForm.jsx
```diff
@@ -0,0 +1,75 @@
...
+      body: 'name=' + this.refs.name.value
+    }).then((res)=>{
+      if(res.ok){
+        return res.json();
+      } else {
+        throw new Error('Unable to create a new app');
+      }
+    }).then((json)=>{
+      if(json.success){
+        let newApp = json.app;
```
wilhelmklopp: this looks like it should be `const`

backend/uclapi/dashboard/api_applications.py
```diff
@@ -43,21 +49,33 @@ def create_app(request):
...
+                "scopes": s.get_all_scopes()
+            }
         }
     })
 
 
 def rename_app(request):
     if request.method != "POST":
-        return HttpResponseBadRequest("Error: Request is not of method POST")
+        return HttpResponseBadRequest(json.dumps({
```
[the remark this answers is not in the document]
ChristopherHammond13: It's probably because the latter will use the pretty printing JSON output. It's more code but I think for DX it's worth it. Opinions?

backend/uclapi/dashboard/api_applications.py
```diff
@@ -43,21 +49,33 @@ def create_app(request):
...
+            "success": False,
+            "error": "Request is not of method POST"
```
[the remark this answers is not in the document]
ChristopherHammond13: any preference? I agree with this.

backend/uclapi/dashboard/api_applications.py
```diff
@@ -183,3 +207,152 @@ def delete_app(request):
...
+        }))
+    try:
+        app_id = request.POST["app_id"]
+        new_callback_url = request.POST["callback_url"]
+        user_id = request.session["user_id"]
+    except KeyError:
+        response = PrettyJsonResponse({
+            "success": False,
+            "message": "Request does not have app_id."
```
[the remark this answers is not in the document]
ChristopherHammond13: Good idea 👍

backend/uclapi/dashboard/api_applications.py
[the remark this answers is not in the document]
wilhelmklopp: Yeah I prefer the latter 👍

backend/uclapi/dashboard/api_applications.py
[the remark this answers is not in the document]
wilhelmklopp: We're using `message` more consistently everywhere I think, but worth double checking that with a grep

backend/uclapi/dashboard/api_applications.py
ChristopherHammond13: This looks quite nice, and I can improve that with the environment variables. If I extract that out into an api_application then when we rebase webhooks on OAuth you could use that function?

backend/uclapi/dashboard/api_applications.py
ChristopherHammond13: 👍

backend/uclapi/dashboard/api_applications.py
ChristopherHammond13: @Faiz triggered

backend/uclapi/dashboard/app_helpers.py
ChristopherHammond13: really? Everywhere I've read online says that the ''.join() is more Pythonic. @Faiz?

backend/uclapi/dashboard/migrations/0008_populate_existing_secrets.py
ChristopherHammond13: Nope! The file naming isn't actually relevant; it's done by the internal graph

backend/uclapi/dashboard/models.py
ChristopherHammond13: Good question. On one hand, I'm okay with doing this because they're fields that are atomic to apps, but on the other hand splitting off OAuth data seems fair too. Depends how separate you see OAuth to the rest of the project.

backend/uclapi/dashboard/urls.py
ChristopherHammond13: These are all internal URLs

backend/uclapi/oauth/models.py
ChristopherHammond13: yeah I know this now

backend/uclapi/oauth/models.py
ChristopherHammond13: nothing, yet ;)

backend/uclapi/oauth/scoping.py
ChristopherHammond13: that should have been done now. Did I not push it?

frontend/src/react/components/appForm.jsx
ChristopherHammond13: This I have no idea about. @hennersz ?

backend/uclapi/oauth/models.py
wilhelmklopp: ok I think in that case we should add that to the model later once we do use it and remove it for now

backend/uclapi/dashboard/urls.py
wilhelmklopp: I think the reason for that may be that we don't have multiple-word urls so far. imo we should have a consistent pattern for that - even for internal URLs. no casing feels like the wrong choice.

backend/uclapi/dashboard/models.py
wilhelmklopp: I think I'd prefer if they were separate purely for flexibility reasons. That's the reason I did it in webhooks if you want to check that out :)

backend/uclapi/dashboard/app_helpers.py
wilhelmklopp: My comment here wasn't about `join` and more about the `+=` concatenation

that could be

```
client_id = '{}.{}'.format(join(str(sr.randint(0, 9)) for _ in range(16)), join(str(sr.randint(0, 9)) for _ in range(16)))
```
not sure what the best f-string syntax would look like cc @jermenkoo

backend/uclapi/dashboard/api_applications.py
wilhelmklopp: yep, that sounds great 💖

wilhelmklopp:
image

There should be some padding above and below the datetimes

wilhelmklopp:
So happy we have this working on our first OAuth app :metal:

image

This is waaaay to busy and verbose I think. Suggested copy and style changes below:
image




GitHub OAuth screen for inspiration:
image

backend/uclapi/dashboard/app_helpers.py
ChristopherHammond13: Whilst your version does look more pythonic, it's far harder to understand and manage. I'm struggling to follow it.

Will consider revising, but we need something that's not about four brackets deep :P

backend/uclapi/dashboard/migrations/0008_populate_existing_secrets.py
ChristopherHammond13: For the record, the reason I kept the 0008 here is that that migration is a multi-step process, but for our purposes, it should be treated as one step. Django just needs a helping hand to pre-populate programmatically generated default values in databases.

backend/uclapi/dashboard/models.py
ChristopherHammond13: Doable: yes
Absolute pain that probably requires wiping out the data in staging and reprogramming the migrations? Also yes.

backend/uclapi/dashboard/urls.py
ChristopherHammond13: We do have one multiple-word URL so far: `user/login.callback`; we use this for the dashboard Shibboleth login callback. The other solution has been to use forward slashes, like `tokens/scopes` and `user/data`

backend/uclapi/oauth/models.py
ChristopherHammond13: wait actually I think when you press deny the token is made inactive

ChristopherHammond13: (no that's not the case, but it should be :P)

backend/uclapi/dashboard/models.py
wilhelmklopp: ¯\_(ツ)_/¯ technical debt
It's definitely easier to change it now than later, but up to you

backend/uclapi/dashboard/migrations/0008_populate_existing_secrets.py
wilhelmklopp: ah ok! just fyi you can also do that in the same migration.

Example: https://github.com/uclapi/uclapi/blob/3a4bf4eac87a93b06d123ce7ffe7c314942ec2a7/backend/uclapi/dashboard/migrations/0008_auto_20170704_2240.py

backend/uclapi/dashboard/urls.py
wilhelmklopp: Can we use one of those here?

backend/uclapi/oauth/models.py
wilhelmklopp: hmm I like the sentiment, but that seems like an antipattern. 🤔 

But for now - does that actually matter? Since an app only gets the standard user data on login right?

backend/uclapi/dashboard/app_helpers.py
wilhelmklopp:
```
client_id = '{}.{}'.format(
    ''.join(str(sr.randint(0, 9)) for _ in range(16)),
    ''.join(str(sr.randint(0, 9)) for _ in range(16))
)
```

ChristopherHammond13: Still about four brackets deep but at least with that indentation it's understandable so I'll accept that change :)

Faiz: sorry late comer. 

Since we are using a function, I would cast the ''.join to a function variable and then call it twice for readability. 

`fun = lambda  : ''.join(str(sr.randint(0, 9)) for _ in range(16))

client_id = '{}.{}'.format(fun(), fun())`

jermenkoo: Named lambdas are pretty much frowned upon, can't you just extract it somewhere, e.g. `utils` as `generate_client_id()` and use that?

Not coupled, testable, replaces 4 lines with 1.

backend/uclapi/oauth/models.py
ChristopherHammond13: I wouldn't call it an anti-pattern _per se_; if I deny an app some privileges I would expect that to trump any acceptance previously given. It deals with an anti-pattern-like scenario, and there's no alternative at the moment.

ChristopherHammond13: I wouldn't call it an anti-pattern _per se_; if I deny an app some privileges I would expect that to trump any acceptance previously given. It deals with an anti-pattern-like scenario, and there's no alternative at the moment.

backend/uclapi/dashboard/app_helpers.py
ChristopherHammond13: Closing this as outdated now: I went with Wilhelm's `.format` syntax. Been using a lot of them at work and it's a nice way of doing things :)

backend/uclapi/dashboard/migrations/0008_populate_existing_secrets.py
ChristopherHammond13: Ah, good to know. I wasn't sure if I could get a guarantee from Django that each of those operations would happen sequentially so I split it into another migration so that it would confirm that the table structure exists before it's populated. I'm fairly new to migrations hence the paranoia :P

backend/uclapi/dashboard/models.py
ChristopherHammond13: Sort of technical debt and sort of isn't. Not having an unnecessary join when every app will have all that data is a good thing. It's a question of balancing DB operations in an app we hope will scale vs. code clarity and avoiding ugly monoliths. I have no particular affinity with either side of this; however, I veer toward leaving it the same because of the effort involved in changing it lol

backend/uclapi/dashboard/api_applications.py
```diff
@@ -1,9 +1,10 @@
-from django.http import HttpResponseBadRequest
+import json
```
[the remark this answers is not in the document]
ChristopherHammond13: This is a bug in Hound from the looks of things as `flake8` returns no errors for this file now

ChristopherHammond13:
I'm going to update staging with this version of the code. Hold tight, ladies and gents!

wilhelmklopp:
@ChristopherHammond13 yep, thanks for pointing that out. Just added the commit to remedy that.

PR #99, opened 2017-09-07
Update setid to LIVE-17-18

wilhelmklopp:
## What does this PR do?
Bumps `setid` to the new academic year 🎓 

## ✅ Pull Request checklist

- [X] Is this code complete?

not applicable

No

## :squirrel: Deploy notes
none


ChristopherHammond13:
I agree that these should be global constants in settings.py. I'll move this value to there then we can merge this.

wilhelmklopp:
Added the global setting. If tests pass should be ready to merge @ChristopherHammond13

PR #100, opened 2017-09-09
Remove dysfunctional button from shibboleth_error

wilhelmklopp:
## What does this PR do?
Removes dysfunctional button that may have been confusing for users

## ✅ Pull Request checklist

- [X] Is this code complete?

not applicable

No

## :squirrel: Deploy notes
Nothing special


ChristopherHammond13:
Whilst we're updating this, shouldn't we have a link to try again or go back to the Dashboard? If you're not bothered (because users should never, under normal circumstances, ever see this screen) then I'm happy to merge.

PR #110, opened 2017-10-03
[Webhook] Re-add lost data migration

wilhelmklopp:
## What does this PR do?
Fixes the data migration that was missing from the webhook PR that got lost in the rebase/re-adding of migrations on top of OAuth migrations

## ✅ Pull Request checklist

- [X] Is this code complete?

No

## :squirrel: Deploy notes
Need to delete the last migration applied on staging (in migrations table)
Should be clean on production

@ChristopherHammond13


ChristopherHammond13:
Tested in production and it works fine, so long as the migration hasn't already been applied (as it's an edit of an existing migration).

PR #115, opened 2017-10-05
Improve room coordinates output

wilhelmklopp:
## What does this PR do?
re-add rooms.json to make local environments more alike and also make sure every room has coordinates, where possible.

## ✅ Pull Request checklist

- [X] Is this code complete?

No

## :squirrel: Deploy notes
Need to run migrations and also `python manage.py insert_locations` as a one off


ChristopherHammond13:
@wilhelmklopp ^^

wilhelmklopp:
@ChristopherHammond13 yes that's correct. Ideally it would only include that.

I'm reusing both an old rooms.json file and an old script. (We've used both previously)

There's not enough time to rewrite this to the ideal case before the hackathon.

ChristopherHammond13:
This looks fine given time constraints, merging.

PR #134, opened 2017-10-27
pc availability app

Faiz:
## What does this PR do?
New service + new endpoint to retrieve PC availability data.

## ✅ Pull Request checklist

- [ ] Is this code complete? Nah, kinda partially.
- [ ] Are tests done/modified?

## 📃 Link to PR updating documentation (if applicable)

No docs as usual

## 🚨 Is this a breaking change for API clients?
No

## :squirrel: Deploy notes
For example: Need to run migrations as part of deployment

## Anything else
Need comments on changes buds

PR #135, opened 2017-10-29
Free rooms

anirudhpillai:
## What does this PR do?
Adds the free rooms endpoint

## ✅ Pull Request checklist

- [ ] Is this code complete?

No

### DIFF
--- a/backend/uclapi/roombookings/helpers.py
+++ b/backend/uclapi/roombookings/helpers.py
@@ -90,6 +90,24 @@ def _localize_time(time_string):
     return ret_time.replace(tzinfo=None)
 
 
+def _round_date(time_string, up=False):
+    """
+    Rounds the datetime to the nearest day.
+    Rounds down by default until up is passed as True in which case
+    it rounds up
+    """
+    date = datetime.date(
+        time_string.year,
+        time_string.month,
+        time_string.day
+    )
+
+    if up:
+        date += datetime.timedelta(days=1)
+
+    return date
+
+
 def _parse_datetime(start_time, end_time, search_date):
     parsed_start_time, parsed_end_time = None, None
     try:
@@ -243,25 +261,47 @@ def _get_bookings_for_room(bookings, siteid, roomid):
     return bookings_for_room
 
 
-def _filter_for_free_rooms(all_rooms, bookings):
+def overlaps(start1, end1, start2, end2):
+    """
+    takes 4 datetimes
+    checks if they overlap
+    """
+    return max(start1, start2) < min(end1, end2)
+
+
+def _filter_for_free_rooms(all_rooms, bookings, start, end):
     """
     Find all rooms which don't have any bookings.
     Args:
-        bookings: All the bookings made in the time period
-        we want to find free rooms in.
         all_rooms: All available rooms.
+        bookings: All the bookings made in the days of the given time period
+        start: Start time for free rooms
+        end: End time for free rooms
     """
-    free_rooms = []
-
+    rooms_with_bookings = list(all_rooms)
     for idx, room in enumerate(all_rooms):
         bookings_for_room = _get_bookings_for_room(
             bookings,
             room["siteid"],
             room["roomid"]
         )
+        rooms_with_bookings[idx]["bookings"] = bookings_for_room
 
-        # Room hasn't been booked in the give time period
-        if not bookings_for_room:
+    free_rooms = []
+    for room in rooms_with_bookings:
+        for booking in room["bookings"]:
+
+            booking_start = _localize_time(booking["start_time"])
+            booking_end = _localize_time(booking["end_time"])
+
+            if overlaps(
+                start1=booking_start,
+                end1=booking_end,
+                start2=start,
+                end2=end
+            ):
+                break
+        else:
             free_rooms.append(room)
 
     return free_rooms
--- a/backend/uclapi/roombookings/helpers.py
+++ b/backend/uclapi/roombookings/helpers.py
@@ -230,3 +230,38 @@ def how_many_seconds_until_midnight():
         day=tomorrow.day, hour=0, minute=0, second=0
     )
     return (midnight - datetime.datetime.now()).seconds
+
+
+def _get_bookings_for_room(bookings, siteid, roomid):
+    bookings_for_room = []
+    for booking in bookings:
+        if (
+            booking["roomid"] == roomid and
+            booking["siteid"] == siteid
+        ):
+            bookings_for_room.append(booking)
+    return bookings_for_room
+
+
+def _filter_for_free_rooms(all_rooms, bookings):
+    """
+    Find all rooms which don't have any bookings.
+    Args:
+        bookings: All the bookings made in the time period
+        we want to find free rooms in.
+        all_rooms: All available rooms.
+    """
+    free_rooms = []
+
+    for idx, room in enumerate(all_rooms):
+        bookings_for_room = _get_bookings_for_room(
+            bookings,
+            room["siteid"],
+            room["roomid"]
+        )
+
+        # Room hasn't been booked in the give time period
+        if not bookings_for_room:
+            free_rooms.append(room)
+
+    return free_rooms
--- a/backend/uclapi/roombookings/urls.py
+++ b/backend/uclapi/roombookings/urls.py
@@ -6,4 +6,5 @@ urlpatterns = [
     url(r'^rooms$', roombookings.views.get_rooms),
     url(r'^bookings$', roombookings.views.get_bookings),
     url(r'^equipment$', roombookings.views.get_equipment),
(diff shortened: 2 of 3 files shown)


ChristopherHammond13:
https://jenkins.uclapi.com/job/UCLAPI/25/

backend/uclapi/roombookings/urls.py
```diff
@@ -6,4 +6,5 @@
     url(r'^rooms$', roombookings.views.get_rooms),
     url(r'^bookings$', roombookings.views.get_bookings),
     url(r'^equipment$', roombookings.views.get_equipment),
+    url(r'^free_rooms$', roombookings.views.free_rooms),
```
wilhelmklopp: I think this one should probably be `freeRooms` to align with our existing endpoints? What do you think @ChristopherHammond13

ChristopherHammond13: @wilhelmklopp which current endpoint is camelCase? I think if any they're just internal APIs which I'm not so concerned about (as we can change / update those without any brekaing changes for clients)

ChristopherHammond13: (I think we should adopt `underscored_endpoints` where possible as it's more Pythonic as per teh conversation about this in #134)

jermenkoo: Should we also name the views like `get_X_Y_Z_`?

backend/uclapi/roombookings/views.py
```diff
@@ -163,3 +164,66 @@ def get_equipment(request, *args, **kwargs):
...
+    # All available rooms
+    all_rooms = Room.objects.using("roombookings").filter(
+        setid=settings.ROOMBOOKINGS_SETID,
+        bookabletype='CB'
+    )
+    all_rooms = _serialize_rooms(all_rooms)
+
+    free_rooms = _filter_for_free_rooms(all_rooms, bookings, start, end)
+
+    return PrettyJsonResponse({
```
jermenkoo: @ChristopherHammond13 here should `kwargs` be passed, right?

backend/uclapi/roombookings/views.py
```diff
@@ -163,3 +164,66 @@ def get_equipment(request, *args, **kwargs):
...
+    # Rounding down start date to start of day
+    request_params["startdatetime__gte"] = _round_date(start)
+
+    # Rounding up end date to start of next day
+    request_params["finishdatetime__lte"] = _round_date(end, up=True)
+
+    # Pagination Logic
+    # maxing out results_per_page to get all the bookings in one page
+    results_per_page = 100000
+    request_params = dict((k, v) for k, v in request_params.items() if v)
```
jermenkoo: You can use a dictionary comprehension directly - `{k:v for k,v in request_params.items() if v}`

backend/uclapi/roombookings/urls.py
ChristopherHammond13: @jermenkoo I'd say so, since we'd do that anyway as it's Pythonic and PEP-8-y

backend/uclapi/roombookings/views.py
ChristopherHammond13: Correct, so that it can extract the rate limit headers and deliver them

backend/uclapi/roombookings/helpers.py
```diff
@@ -230,3 +248,60 @@ def how_many_seconds_until_midnight():
         day=tomorrow.day, hour=0, minute=0, second=0
     )
     return (midnight - datetime.datetime.now()).seconds
+
+
+def _get_bookings_for_room(bookings, siteid, roomid):
```
ChristopherHammond13: Can we not do this with a database query? It should be far more efficient (e.g. use the cache lock then do `bookings = BookingX.objects.filter(siteid=siteid, roomid=roomid)` (where X = A or B depending on the current state of the lock) then use our existing code to turn this into JSON. It should be far more efficient...

backend/uclapi/roombookings/helpers.py
```diff
@@ -230,3 +248,60 @@ def how_many_seconds_until_midnight():
...
+            room["siteid"],
+            room["roomid"]
+        )
+        rooms_with_bookings[idx]["bookings"] = bookings_for_room
+
+    free_rooms = []
+    for room in rooms_with_bookings:
+        for booking in room["bookings"]:
+
+            booking_start = _localize_time(booking["start_time"])
```
ChristopherHammond13: This could probably be done in `_get_bookings_for_room` to avoid looping over all the data again (unless I misunderstood something, which is entirely possible)

backend/uclapi/roombookings/helpers.py
```diff
@@ -230,3 +248,60 @@ def how_many_seconds_until_midnight():
...
+            booking_end = _localize_time(booking["end_time"])
```
ChristopherHammond13: as above

backend/uclapi/roombookings/views.py
```diff
@@ -163,3 +164,66 @@ def get_equipment(request, *args, **kwargs):
...
+    request_params['finishdatetime__lte'] = request.GET.get('end_datetime')
+
+    if (
+        not request_params['startdatetime__gte'] or
+        not request_params['finishdatetime__lte']
+    ):
+        return PrettyJsonResponse({
+            "ok": False,
+            "error": "start_datetime or end_datetime not provided"
+        })
```
ChristopherHammond13: The rate limit will still have been applied so we should pass `kwargs` to the `PrettyJsonResponse` too so that the headers get sent to the client.

backend/uclapi/roombookings/views.py
```diff
@@ -163,3 +164,66 @@ def get_equipment(request, *args, **kwargs):
...
+        request_params['startdatetime__gte'],
+        request_params['finishdatetime__lte'],
+        None
+    )
+
+    if not is_parsed:
+        return PrettyJsonResponse({
+            "ok": False,
+            "error": "date/time isn't formatted as suggested in the docs"
+        })
```
ChristopherHammond13: Again, `kwargs` need to be passed here.

backend/uclapi/roombookings/views.py
```diff
@@ -163,3 +164,66 @@ def get_equipment(request, *args, **kwargs):
...
+
+    # Rounding down start date to start of day
+    request_params["startdatetime__gte"] = _round_date(start)
+
+    # Rounding up end date to start of next day
+    request_params["finishdatetime__lte"] = _round_date(end, up=True)
+
+    # Pagination Logic
+    # maxing out results_per_page to get all the bookings in one page
+    results_per_page = 100000
```
ChristopherHammond13: The number of bookings in the database is approx. 240,000. Is this going to be an issue?

backend/uclapi/roombookings/urls.py
wilhelmklopp: > @wilhelmklopp which current endpoint is camelCase?

I think you're right we're using flat case [(described here)](https://stackoverflow.com/questions/17326185/what-are-the-different-kinds-of-cases)

> (I think we should adopt underscored_endpoints where possible as it's more Pythonic as per teh conversation about this in #134)

disagree on this one because the endpoints aren't in python (talked about this in person today)
Our function/view names should definitely be pythonic

ChristopherHammond13:
https://jenkins.uclapi.com/job/UCLAPI/31/

ChristopherHammond13:
@anirudhpillai 
- Is this ready for merging now once the conflict is resolved?
- Is it documented in the new docs?

ChristopherHammond13:
https://jenkins.uclapi.com/job/UCLAPI/43/

ChristopherHammond13:
https://jenkins.uclapi.com/job/UCLAPI/50/

ChristopherHammond13:
https://jenkins.uclapi.com/job/UCLAPI/106/

ChristopherHammond13:
https://jenkins.uclapi.com/job/UCLAPI/149/

backend/uclapi/roombookings/views.py
```diff
@@ -163,3 +164,66 @@ def get_equipment(request, *args, **kwargs):
...
```
wilhelmklopp: @anirudhpillai I don't fully understand this line. Is the idea just that we skip pagination entirely?

anirudhpillai: Yeah, we're trying to get everything in one page.

anirudhpillai: Yes, that's what it is doing. This lets us reuse the caching function written by Faiz. Also, it shouldn't actually return those many bookings as we only look up bookings between the given start and end date.

wilhelmklopp:
@ChristopherHammond13 alright, let's merge this 👍

ChristopherHammond13:
https://jenkins.uclapi.com/job/UCLAPI/213/

PR #136, opened 2017-10-31
restart autoupdate counter on the database from django itself

Faiz:
## What does this PR do?
This executes a raw query everytime buckets are refilled to ensure that the auto increment starts from 0 for the primary key ids. This also replaces the `Model.objects.all().delete()` with truncate. Using raw query should give some millisecond advantage.

## ✅ Pull Request checklist

- [x] Is this code complete?

### DIFF
--- a/backend/uclapi/roombookings/management/commands/update_gencache.py
+++ b/backend/uclapi/roombookings/management/commands/update_gencache.py
@@ -2,6 +2,7 @@ import os
 
 from django.core.management.base import BaseCommand
 from django.conf import settings
+from django.db import connections
 
 import cx_Oracle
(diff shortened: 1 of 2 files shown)


ChristopherHammond13:
https://jenkins.uclapi.com/job/UCLAPI/29/

wilhelmklopp:
I don't think I know enough, so I'll leave this to Chris

@Faiz have you tried running this? Does it work?

And this definitely resets the sequence too, right, it doesn't just truncate the table?

Faiz:
I have run it 4 times to make it works. You can run it in staging as well to see if I am right.

PR #182, opened 2017-11-16
Change docs references to objects, dicts and maps 

keelanfh:
## What does this PR do?
The current docs refer to this JSON structure:
`{"foo": "bar"}`
as variously an object, a dict, and even a map. 

This is confusing when reading the docs, so I propose changing it to make it consistent.
Per the [JSON spec](http://json.org/), this is called an object, so I've gone with that, although I think the most important thing is that the wording is consistent. 

Would be interesting to see what you think :)

## ✅ Pull Request checklist

- [ ] Is this code complete?

Yes/No

## :squirrel: Deploy notes


ChristopherHammond13:
Build finished. 

ChristopherHammond13:
https://jenkins.uclapi.com/job/UCLAPI/111/

PR #187, opened 2017-11-22
Fix "undefined" in webhook settings

wilhelmklopp:
## What does this PR do?
Fixes a bug where we'd store "undefined" in the db after a user entered no value.

Also makes the input text a different color to the placeholder, which was causing problems previously.

Also adds the package-lock.json file so that we can make sure we all have the same dependency installed and enable dependabot ✅ 

## ✅ Pull Request checklist

- [X] Is this code complete?

No

image

### DIFF
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -5,7 +5,6 @@
         "node": true
     },
     "extends": "eslint:recommended",
-    "installedESLint": true,
     "parserOptions": {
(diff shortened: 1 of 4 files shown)


ChristopherHammond13:
Build finished. 

ChristopherHammond13:
https://jenkins.uclapi.com/job/UCLAPI/122/

PR #248, opened 2017-12-22
Gencache Work

ChristopherHammond13:
## What does this PR do?
Updates the `update_gencache` script to be more RAM efficient, and adds a cron script that we can finally manage using Git so that we don't have to run `nano` on prod servers (awks).

## ✅ Pull Request checklist

- [x] Is this code complete?

No

## :squirrel: Deploy notes
Deploy to a new cron box, point the script at a new RDS instance and let things settle. We can finally sunset the RAM Cache after this...


ChristopherHammond13:
https://jenkins.uclapi.com/job/UCLAPI/382/

ChristopherHammond13:
https://jenkins.uclapi.com/job/UCLAPI/383/

backend/uclapi/roombookings/management/commands/update_gencache.py
```diff
@@ -35,53 +36,70 @@ def handle(self, *args, **options):
 
         self.stdout.write("Selecting a clone table...")
         lock = Lock.objects.all()[0]
-        curr = BookingA if lock.bookingA else BookingB
+        current_bucket = BookingA if lock.bookingA else BookingB
```
wilhelmklopp: I applaud this name change 👏 I always got confused with what `curr` meant

backend/uclapi/roombookings/management/commands/update_gencache.py
```diff
@@ -35,53 +36,70 @@ def handle(self, *args, **options):
...
-            batch_size=5000
-        )
+
+        batch_size = 5000
+        running_total = 0
+
+        while True:
+            data_objects = []
+            for row in cur.fetchmany(numRows=batch_size):
+                data_objects.append(current_bucket(
```
jermenkoo: Can't this be `current_bucket(*row)`? @Faiz

backend/uclapi/roombookings/management/commands/update_gencache.py
ChristopherHammond13: Yeah I think we were due a bit of a refactor of this function :P

backend/uclapi/roombookings/management/commands/update_gencache.py
ChristopherHammond13: I'm not sure it can be because we match the fields to Django model definitions. Plus, why `*row` and not `**row`? Perhaps that my Python knowledge lacking...

ChristopherHammond13:
https://jenkins.uclapi.com/job/UCLAPI/419/

backend/uclapi/roombookings/management/commands/update_gencache.py
jermenkoo: > we match the fields to Django model definitions

True, have not considered that. (I thought the model definition has it in the order we specify it)

> why `*row` and not `**row`? 

`**` should be used with mappings (e.g. dict); `fetchmany(...)` returns a tuple which you access using indices therefore `*`.

ChristopherHammond13: My only concern is that if the underlying table or view structure changes (even with the addition of just one more field) the code will error out. At the moment we are fine unless data is removed, but if data is added nothing will break. If that changes in the future (e.g. if UCL do another join in our view's code) then doing `**row` will break the caching.

ChristopherHammond13: (I love the Pythonic magic of the double asterisk though! I'm very conflicted on this)

ChristopherHammond13:
https://jenkins.uclapi.com/job/UCLAPI/460/

PR #864, opened 2018-12-27
added client_secret as required query parameter to Timetable API calls

skabra07:
## What does this PR do?
Current documentation did not include client_secret as a required query parameter for Timetable API calls. Hence I added that, updated the example code and also added the error responses if client_secret is not provided or is incorrect.

## ✅ Pull Request checklist

- [x] Is this code complete?

No

### DIFF
--- a/frontend/src/components/documentation/Routes/Timetable/GetDataDepartments.jsx
+++ b/frontend/src/components/documentation/Routes/Timetable/GetDataDepartments.jsx
@@ -9,15 +9,17 @@ let codeExamples = {
   python: `import requests
 
 params = {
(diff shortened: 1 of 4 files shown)


skabra07:
Just tested it. As you said, using the application token did not require the client secret but using the user generated token did. I guess just updating that in your docs (differentiating between user generated token and application token) would be useful. Thanks again!

ChristopherHammond13:
> 
> 
> Just tested it. As you said, using the application token did not require the client secret but using the user generated token did. I guess just updating that in your docs (differentiating between user generated token and application token) would be useful. Thanks again!

I'm glad that worked for you! I'm going to close this PR, but we'll open an issue to make sure we properly explain the types of tokens more clearly to users. Thanks for raising this! 🎉

PR #935, opened 2019-01-23
Expires Header WIP

### LINKED ISSUE #839: Cached Data Age Headers
We have a lot of data that is updated regularly from other sources and cached at our end. We should tell our developers programmatically how old this data is so that they can plan their apps around our updating.

This will also help us to build #838 because if data is too old then we know that we have an issue with our Cron server (i.e. the data is not being updated as expected).

We can use the [Expires](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires) header to implement this.

zipy124:
## What does this PR do?
Start of work on Expires header for the appropriate methods, currently implemented on /bookings as an example. Time is stored in time.txt for now(this is bad, maybe we should add a model for time when cache was updated?) stores cache time plus 20mins when running update_gencache.

## ✅ Pull Request checklist

- [No] Is this code complete?
- [No] Are tests done/modified? (Should probably write a test for this)
- [No] Are docs written for this PR? (Definitely should mention this in docs)

No

## :squirrel: Deploy notes
No Deploy notes that I can think of?

## Anything else
Closes #839 when finished

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,3 +133,6 @@ dump.rdb
 /scripts/shibboleth/sp-cert.staging.pem
 /scripts/shibboleth/sp-key.prod.pem
 /scripts/shibboleth/sp-key.staging.pem
+
+# Time cache file
+time.txt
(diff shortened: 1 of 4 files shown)


ChristopherHammond13:
So the main issue with storing the time in a file is that the file won't exist on every web server (in fact, it would only exist on the cron box that does the caching).

I think putting the last update time into Redis makes most sense here?
This might also be a better header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified

zipy124:
> So the main issue with storing the time in a file is that the file won't exist on every web server (in fact, it would only exist on the cron box that does the caching).
> 
> I think putting the last update time into Redis makes most sense here?
> This might also be a better header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified

yep, i'll abandon this branch

PR #940, opened 2019-01-24
Redis-based Last-Modified Header

### LINKED ISSUE #839: Cached Data Age Headers

ChristopherHammond13:
## What does this PR do?
Adds Last-Modified Header

## ✅ Pull Request checklist

- [x] Is this code complete?

No

## :squirrel: Deploy notes
Just deploy!

## Anything else
closes #839

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,6 +84,9 @@ celerybeat-schedule
 # dotenv
 .env
 
+# Other .env files
+*.env
+
 # virtualenv
 venv/
 ENV/
--- a/backend/uclapi/common/decorators.py
+++ b/backend/uclapi/common/decorators.py
@@ -269,6 +269,7 @@ def _get_last_modified_header(redis_key=None):
 
     return last_modified
 
+
 def uclapi_protected_endpoint(
     personal_data=False,
     required_scopes=[],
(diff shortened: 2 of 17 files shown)


zipy124:
Seems good to me. I'm not sure the testing involving current_time is correct? We assert in the comments that we can't be sure it executes within the second, therefore the test could fail if we execute at say 16:59:59 and the other timestamp is 16:60:00. the seconds are technically not within 1 of another, maybe add a modulo to wrap around at 60 and 00?

zipy124:
@ChristopherHammond13 We should probably add some information about the header details to the documentation? Where shall we put it?

zipy124:
@ChristopherHammond13 Timetable events draw from both the timetable_gencache and gencache(through the _get_timetable_events function). Which header should they therefore take?

zipy124:
> Good point. We can probably do it by generating all datetimes for within 3 seconds of the current one and asserting that they're within range. We'd hit the same issue if we executed over midnight, for example.

Sounds like a good solution to me.

ChristopherHammond13:
I'll merge this once I've fixed up the test as per how we discussed it on Slack @zipy124

PR #952, opened 2019-01-26
Removes seemingly unnecessary dotenv import in common.helpers

ChristopherHammond13:
## What does this PR do?
Should get rid of the annoying error when running the API

## ✅ Pull Request checklist

- [x] Is this code complete?

No

## :squirrel: Deploy notes
Make sure this doesn't break anything and merge away!

### DIFF
--- a/backend/uclapi/common/helpers.py
+++ b/backend/uclapi/common/helpers.py
@@ -4,12 +4,6 @@ import textwrap
 from binascii import hexlify
 
 from django.http import JsonResponse, HttpResponse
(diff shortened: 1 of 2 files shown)


zipy124:
Should have fixed this, you removed all .env imports instead of the times we actually import it with no parameters, I simply removed the import call for common/.env at the beginning of manage.py

ChristopherHammond13:
@zipy124 Is this definitely the fix? It looks like we have a useless import now and no `read_dotenv` in `manage.py`. Won't this break development?

PR #222, opened 2017-12-05
Temp token to rediis

Faiz:
## What does this PR do?
Moves temporary token to redis

## ✅ Pull Request checklist

- [x ] Is this code complete?

## 🚨 Is this a breaking change for API clients?
Yes

## :squirrel: Deploy notes
Run migration.

PR #960, opened 2019-01-27
Enabled and fixed temp token tests. Add common/helpers.py tests

zipy124:
## What does this PR do?
Enabled and fixed temp-token tests
Also Adds test to see if headers are added correctly to http responses.
## ✅ Pull Request checklist

- [ ] Is this code complete?

No

## :squirrel: Deploy notes
Run the new migration to delete the un-needed model


ChristopherHammond13:
Great code!
Just a heads up though that the `TemporaryToken` model is now no longer used, and we instead put temp tokens in Redis to avoid filling up the database (this was done in #222, which was only merged and completed very recently).

It might be worth us stripping out all the code for the `TemporaryToken` model, but for now I'm happy to keep tests in place for it.

ChristopherHammond13:
(In fact, if you want to strip all that legacy code out in this PR feel free, because then there'll be less code to test too)

zipy124:
> (In fact, if you want to strip all that legacy code out in this PR feel free, because then there'll be less code to test too)

I suspected there was a reason for this, i was very confused. I'll try to strip out all the legacy code and their tests!

zipy124:
@ChristopherHammond13 I've stripped out the legacy model. The tests for expired tokens has also been removed temporarily while i re-make it work with redis.

PR #964, opened 2019-01-27
Fix for webpack-stats.json deployment

hueyy:
## What does this PR do?
* `webpack-stats.json` now generated in `dashboard/statics`
* `wsgi.py` now downloads `webpack-stats.json` from S3 upon run

## ✅ Pull Request checklist

- [x] Is this code complete?

No

## :squirrel: Deploy notes
* `collectstatic` should upload `webpack-stats.json` to S3 bucket

### DIFF
--- a/backend/uclapi/uclapi/settings.py
+++ b/backend/uclapi/uclapi/settings.py
@@ -215,7 +215,10 @@ WEBPACK_LOADER = {
     'DEFAULT': {
         'CACHE': not DEBUG,
         'BUNDLE_DIR_NAME': '/',  # must end with slash
-        'STATS_FILE': os.path.join(BASE_DIR, 'dashboard/static/webpack-stats.json'),
+        'STATS_FILE': os.path.join(
+            BASE_DIR,
+            'static/webpack-stats.json'
+        ),
         'POLL_INTERVAL': 0.1,
         'TIMEOUT': None,
         'IGNORE': [r'.+\.hot-update.js', r'.+\.map']
@@ -242,7 +245,7 @@ WEBPACK_LOADER = {
 #    of a person who has permission to upload new statics to
 #    S3.
 
-if strtobool(os.environ.get("AWS_S3_STATICS", "False")) or True:
+if strtobool(os.environ.get("AWS_S3_STATICS", "False")):
     DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
     STATICFILES_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
     AWS_STORAGE_BUCKET_NAME = os.environ["AWS_S3_BUCKET_NAME"]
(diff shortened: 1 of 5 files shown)


backend/uclapi/uclapi/settings.py
```diff
@@ -242,7 +242,7 @@
 #    of a person who has permission to upload new statics to
 #    S3.
 
-if strtobool(os.environ.get("AWS_S3_STATICS", "False")):
+if strtobool(os.environ.get("AWS_S3_STATICS", "False")) or True:
```
ChristopherHammond13: I assume this line was put in for debug purposes? We don't use AWS S3 statics locally, only in prod / staging.

backend/uclapi/uclapi/wsgi.py
```diff
@@ -28,4 +30,11 @@
 if os.environ.get("EVENTLET_NOPATCH") != 'True':
     eventlet.monkey_patch()
 
+# Download latest webpack-stats.json from S3
+# before starting WSGI
+WEBPACK_STATS_URL = "https://{}/static/webpack-stats.json".format(settings.AWS_S3_CUSTOM_DOMAIN)
```
ChristopherHammond13: There is another slight issue: we run the code in parallel so it's possible that one of the threads might attempt to download the file before the others and they'll end up trampling on each other. Perhaps we need to check the file age or see if there's a lock on it before attempting to overwrite it? Otherwise we'll get nasty concurrency issues in prod.

I think at any one time we have about four Gunicorn workers per host.

backend/uclapi/uclapi/settings.py
hueyy: yup yup oops

ChristopherHammond13:
Also this: https://travis-ci.org/uclapi/uclapi/builds/485083434#L2234
I assume that's easy to fix though.

hueyy:
Right now I've got this:


```python
# check if another gunicorn process is
# already writing to it
try:
    fcntl.flock(WEBPACK_STATS_LOC, fcntl.LOCK_EX | fcntl.LOCK_NB)
    with urllib.request.urlopen(WEBPACK_STATS_URL) as response:
        with open(WEBPACK_STATS_LOC, 'wb') as out_file:
            shutil.copyfileobj(response, out_file)
            fcntl.flock(WEBPACK_STATS_LOC, fcntl.LOCK_UN)
except IOError as e:
    if e.errno != errno.EAGAIN:
        raise
    pass
```

This attempts to lock the file before downloading a new one. If it can't get a lock, one of the other gunicorn processes must already be in the process of downloading & saving the file so all's well. This seems okay?

ChristopherHammond13:
> 
> 
> Right now I've got this:
> 
> ```python
> # check if another gunicorn process is
> # already writing to it
> try:
>     fcntl.flock(WEBPACK_STATS_LOC, fcntl.LOCK_EX | fcntl.LOCK_NB)
>     with urllib.request.urlopen(WEBPACK_STATS_URL) as response:
>         with open(WEBPACK_STATS_LOC, 'wb') as out_file:
>             shutil.copyfileobj(response, out_file)
>             fcntl.flock(WEBPACK_STATS_LOC, fcntl.LOCK_UN)
> except IOError as e:
>     if e.errno != errno.EAGAIN:
>         raise
>     pass
> ```
> 
> This attempts to lock the file before downloading a new one. If it can't get a lock, one of the other gunicorn processes must already be in the process of downloading & saving the file so all's well. This seems okay?

That looks good to me! Good job

Issue #968, opened 2019-01-27
Restore Missing Frontend Assets


ChristopherHammond13:
Since #916 the UCL API logo doesn't show on the frontend, and some docs statics don't render either.

This regression would not have made it into production, but since production was essentially down due to the issues we had before, I decided to take the regression so that at least the service would be functional and alive again.

![image](https://user-images.githubusercontent.com/295396/51808529-edafd480-228c-11e9-8407-d846576e07a9.png)
![image](https://user-images.githubusercontent.com/295396/51808531-f9030000-228c-11e9-9791-c96552493b04.png)

hueyy:
The issue here is django-webpack-loader generates the path for JS/CSS whilst webpack generates the relative path for images & other files.

Webpack generates a relative path (currently `./`). Instead, it should generate a path to the S3 asset. This can be addressed by configuring an `output.publicPath`. However, this publicPath will differ between development & production environments (between `./static/` and `https://uclapi-static.s3.amazonaws.com/static/`)

The most logical way to achieve this seems to be to read the environment variables in (like the backend does) and generate the appropriate publicPath.

PR #969, opened 2019-01-28
Restore missing frontend assets

hueyy:
## What does this PR do?
* Restores missing frontend assets (#968)
* Fixes the paths for images, etc.
* Sets `output.publicPath` in Webpack based on environment variables

## ✅ Pull Request checklist

- [x] Is this code complete?

No

## :squirrel: Deploy notes
* Need to re-upload static assets to S3


hueyy:
Wait, I just realised we might not use `.env` files in non-dev environments. But this is okay because we build frontend assets on our own machines which have `.env` files. And even if the environment variables are loaded w/o a `.env` file, the import of the `.env` file will silently fail and the loaded environment variables are still used. Okay, seems fine.

frontend/package.json
```diff
@@ -22,6 +22,7 @@
   "dependencies": {
     "ajv": "^6.7.0",
     "aphrodite": "^2.2.3",
+    "dotenv": "^6.2.0",
```
ChristopherHammond13: Would this actually be a dev dependency as opposed to a general one?

frontend/webpack.dev.js
```diff
@@ -1,10 +1,19 @@
+require('dotenv')
```
ChristopherHammond13: Do we actually need this logic in `webpack.dev.js`? I'm not sure there's a situation where we'd want to build dev assets for prod, unless I've missed something of course.

frontend/package.json
hueyy: Fair point.

frontend/webpack.dev.js
hueyy: That makes a lot of sense.

PR #976, opened 2019-01-29
Adds connection pooling

### LINKED ISSUE #920: Scaling of PostgreSQL Connections
We have tried in the past to implement connection pooling for Django <==> PostgreSQL. We should try this again so that the system doesn't get pulled down by too many parallel requests (as could happen in the event of UCL Assistant).

zipy124:
## What does this PR do?
Adds connection pooling

## ✅ Pull Request checklist

- [x] Is this code complete?

No

## :squirrel: Deploy notes
We should set pool size to the limit of our postgresql backend, i don't know how many it allows by default.

## Anything else
closes #920


ChristopherHammond13:
This looks good! I'm not going to merge until we have staging to load test this though :)

PR #977, opened 2019-01-29
Webhook fix (Come at me hound)

### LINKED ISSUE #527: Webhooks
They haven't been delivered for a _long_ time now, so we finally need to dig deep and see what is going wrong. The room data is definitely up to date, however.

zipy124:
## What does this PR do?
WEBHOOKS WORK AGAIN?

## ✅ Pull Request checklist

- [ ] Is this code complete?

No

## :squirrel: Deploy notes
nothing of note? update the cron box with this i guess. --debug flag can be set if you manually run the webhook task to get status codes on the responses.

## Anything else
closes #527


backend/uclapi/roombookings/management/commands/trigger_webhooks.py
```diff
@@ -121,18 +130,23 @@ def webhook_filter(booking):
...
+                        headers={
                             "User-Agent": "uclapi-bot/1"
                         }
                     )
                 )
         self.stdout.write(
             "Triggering {} webhooks.".format(len(unsent_requests))
         )
-        grequests.map(unsent_requests)
+        if("debug" in options):
```
ChristopherHammond13: Where is this set?

backend/uclapi/requirements.txt
```diff
@@ -41,6 +41,7 @@ raven==6.10.0
 redis==3.0.1
 requests==2.21.0
 requests-file==1.4.3
+requests-futures==0.9.9
```
ChristopherHammond13: No problem with adding a requirement. But isn't there a previous one that we can remove now?

backend/uclapi/roombookings/management/commands/trigger_webhooks.py
zipy124: its a flag you can pass to the command --debug, when running individually instead of via the gencache. It just enables easier debugging in the future if we notice this takes a long time.

backend/uclapi/requirements.txt
zipy124: Yes, we can remove grequests i believe, let me modify this list

backend/uclapi/roombookings/management/commands/trigger_webhooks.py
ChristopherHammond13: fantastic, thanks!

PR #996, opened 2019-02-05
Migration of /rooms to postgres, unifies Lock

### LINKED ISSUE #1067: Unify Locking
At the moment there are two `gencache` locks: one for roombookings, and one for timetable. We should unify this such that we have one lock shared between the two, so that we can guarantee correctness when we combine data from the two. This will be useful in #1064.

zipy124:
## What does this PR do?
Migrates /rooms to postgres. Currently only the automated response left to do. Unifies Lock and caching

## ✅ Pull Request checklist

- [ ] Is this code complete?

No

## :squirrel: Deploy notes
make and run migrations, change caching script to only call the one caching script now.

## Anything else
Closes #1067


ChristopherHammond13:
We need this prod-ified soon please so that #1064 can be updated to support the new locking 💖

zipy124:
> We need this prod-ified soon please so that #1064 can be updated to support the new locking sparkling_heart

@ChristopherHammond13 I'll hound fix it now. And also add the script to delete all but 1 locks! should be ready to push by the end of today.

zipy124:
@ChristopherHammond13 33 hound violations are in-regard to models. Do i leave these lines or split them up? I don't think it makes sense to multi-line them personally. Everything we discussed is implemented. Webhooks might not work on the first cache, but after that they'll be synced again. I'm going to run the caching script a few times in uni today to make sure it works and also to see if I can optimize the speed at all.

ChristopherHammond13:
This is now handled in #1064

PR #1018, opened 2019-02-11
New Dev Docker Environment

ChristopherHammond13:
## What does this PR do?
New Docker development environment so that all development can be done under Docker containers. It will also automatically provision `fakeshibboleth` and most of `django-uclapi-oauth`.

## ✅ Pull Request checklist

- [ ] Is this code complete?

No

## :squirrel: Deploy notes
Just run `docker-compose up` to get the environment online, and browse to `http://localhost:8000` for the API.


ChristopherHammond13:
A couple of my own comments / observation on this PR:
- The `.env` solution to get the current working directory is a hacky one. It might not even be needed now, but it very much depends on the Docker Compose context parsing. I'll test this one out.
- I'm not sure whether creating a lock on every start is necessary, or if that should be part of another set up script (which includes doing an actual gencache)
- The environment variables for `django-uclapi-oauth`'s `UCLAPI_CLIENT_ID` and `UCLAPI_CLIENT_SECRET` can't yet be filled in. One solution would be to code a Django Management Command that sets up the development user in Django with a dev app and dev token that's well known, and then also configure this value in the `django-uclapi-oauth` system. It's not ideal, and it feels like a lot of extra work, but it could do the trick?

Issue #983, opened 2019-02-01
Duplicate Timetable Entries


Feverfew:
Currently there is a bug where there are multiple timetable entries for the same event, due to it being booked for different people (lecturer/module leader etc.).

PR #1064, opened 2019-02-21
[WIP] Migration of Timetable Generation to Stored Function

### LINKED ISSUE #467: Timetable Performance
We need to speed Timetable up by quite a bit.

ChristopherHammond13:
## What does this PR do?
Will migrate all timetable generation to the RDS Cache so that web servers do not have to load masses of Timetable data into RAM.

## ✅ Pull Request checklist

- [ ] Is this code complete?

No

## :squirrel: Deploy notes
Unsure as of yet. Migrations and caching will definitely be required, and then we'll need a way to install the function into Postgres. It may be possible to use Django to do this.

## Anything else
Resolves #467 and #983 once and for all.


ChristopherHammond13:
I'm thinking we should merge #996 into this branch since it also involves migrations?

ChristopherHammond13:
This will be ready to merge very soon. I just need to finish off my new `update_gencache` code to show progress across the multiprocessing

PR #1090, opened 2019-02-26
Add verbosity to callback url errors

### LINKED ISSUE #1087: Add verbosity to oauth callback URL error message
We should tell users whats wrong with their url, for example not starting with https:// etc...

zipy124:
## What does this PR do?
Adds a reason for why the oauth callback url is not valid

## ✅ Pull Request checklist

- [ ] Is this code complete?

no

## :squirrel: Deploy notes
n/a

## Anything else
closes #1087


zipy124:
example screenshots:
screenshot from 2019-02-26 23-24-09
screenshot from 2019-02-26 23-24-21
screenshot from 2019-02-26 23-24-31

zipy124:
> This code in principle is excellent, but I'd much prefer if we had some defined constants in place of the integer response codes if possible? I think it'll be a bit less clunky that way, and easier to maintain in the future. Also, `is_url_unsafe` is probably a better function name :P

Can't believe I didn't call the function that tbh :laughing: @ChristopherHammond13 Where do i define the constants? in the settings file?

ChristopherHammond13:
@zipy124 you can define constants in the script where the function is declared, then import them at the same time as the function is imported :) keeps the settings file more chill

zipy124:
@ChristopherHammond13 Constants are added and all work corerctly. Maybe you want to re-name them i'm not the best namer of things.

backend/uclapi/dashboard/api_applications.py
```diff
@@ -251,12 +252,20 @@ def set_callback_url(request):
...
+    url_not_safe_saved = is_url_unsafe(new_callback_url)
+    if url_not_safe_saved:
+        if url_not_safe_saved == NOT_HTTPS:
+            message = "The requested callback URL does not "\
+                      "start with 'https://'."
+        elif url_not_safe_saved == NOT_VALID:
+            message = "The requested callback URL is not valid."
+        elif url_not_safe_saved == URL_BLACKLISTED:
+            message = "The requested callback URL is forbidden."
+        elif url_not_safe_saved == NOT_PUBLIC:
```
ChristopherHammond13: Is there ever a case when none of these clauses could be called? Is it worth having a generic fall through `else` that just says that there's an unknown error with setting the callback URL?

backend/uclapi/dashboard/api_applications.py
```diff
@@ -251,12 +252,20 @@ def set_callback_url(request):
...
+            message = "The requested callback URL is not publicly available."
         response = PrettyJsonResponse({
             "success": False,
-            "message": ("The requested callback URL"
-                        " is not valid.")
+            "message": (message)
```
ChristopherHammond13: Are those brackets needed?
```suggestion
            "message": message
```

backend/uclapi/dashboard/api_applications.py
zipy124: no other error scan occur

PR #1131, opened 2019-03-12
[Workspaces] Allow Developers to Filter Out Surveys

### LINKED ISSUE #1025: Workspaces API should show which spaces are actually available to students
At the moment we return spaces such as 1SMLG, which is actually an ISD office space and not a student study area. We should curate this and allow users to specify if they want to enable non-student study areas or not.

ChristopherHammond13:
## What does this PR do?
Allows users to filter out surveys.
E.g.
- `survey_filter=all`: return all surveys
- `survey_filter=student`: return only surveys designated as student study areas
- `survey_filter=staff`: return only surveys designated as staff study areas

This new parameter applies to `/surveys`, `/sensors/summary` and `/sensors/averages/time`.

We define in code which survey IDs are which.

This PR also does the following:
- Fixes up our full OccupEye cache script to discard old surveys if they have an `EndDate` attribute which is in the past
- Adds some test cases for `str2bool`
- Returns in `/surveys` whether each workspace is for staff or not

These changes are now high priority to support https://github.com/uclapi/ucl-assistant-app/issues/77
This PR will also resolve #1025 
## ✅ Pull Request checklist

- [ ] Is this code complete?

YES: We will now no longer return staff-only workspaces by default. API Clients wishing to access *staff* study spaces **must** now pass `survey_filter=all` or `survey_filter=staff` to obtain this data in summaries. Endpoints where the `survey_id` is passed directly work as before but if the `survey_filter` parameter is in place and staff workspaces are required then it must be provided, even if `survey_ids` is used.

## :squirrel: Deploy notes
Run Full OccupEye Cache.


backend/uclapi/workspaces/occupeye/utils.py
```diff
@@ -31,7 +31,12 @@ def authenticated_request(url, bearer):
 def survey_ids_to_surveys(surveys_data, survey_ids):
         if survey_ids:
             try:
-                survey_ids_list = [int(x) for x in survey_ids.split(',')]
+                # If we cast to a set then a list we remove duplicates
+                # This avoids the user specifying a filter with a bad
+                # ID and a duplicate good one, which would cause issues.
+                survey_ids_list = list(
+                    set([int(x) for x in survey_ids.split(',')])
+                )
```
Feverfew: Is there any particular reason to convert back to a list? We're storing hashable elements (ints) and not using any list specific functions from what I can tell so shouldn't hit into runtime errors.

Feverfew:
I was wondering, why do we not use distutils.util.strtobool (https://docs.python.org/3/distutils/apiref.html) instead of our own implementation? For a start it wouldn't return false on strings such as 'maybe', instead it would call ValueError, which I believe is a bit more 'explicit'. I'm an advocate of using the standard library where possible.

Also, I may be wrong but I believe the distinction between study space access is a bit more granular. For example, as an undergraduate I am not allowed in the Graduate Hub. I've also been told (don't know for sure) that I am not allowed in the Bartlett Library as I am not an architecture student. I believe shibboleth does provide some group (i.e. department) information so if we can offer this granularity I think we should. We should probably think about this now before merging as if we do find out we can be more granular and we decide to offer it; it might constitute a breaking change depending on preferred implementation of a more detailed filter.

Otherwise, I'm happy with the code itself.

backend/uclapi/workspaces/occupeye/utils.py
ChristopherHammond13: Good question. The reason was because it was a list to start with. If we can use sets in the same way as lists then I'm happy to remove that.

ChristopherHammond13:
> I was wondering, why do we not use distutils.util.strtobool (https://docs.python.org/3/distutils/apiref.html) instead of our own implementation? For a start it wouldn't return false on strings such as 'maybe', instead it would call ValueError, which I believe is a bit more 'explicit'. I'm an advocate of using the standard library where possible.
We have used our own str2bool implementation for a while, but you're right that we should deprecate that.


> Also, I may be wrong but I believe the distinction between study space access is a bit more granular. For example, as an undergraduate I am not allowed in the Graduate Hub. I've also been told (don't know for sure) that I am not allowed in the Bartlett Library as I am not an architecture student. I believe shibboleth does provide some group (i.e. department) information so if we can offer this granularity I think we should. We should probably think about this now before merging as if we do find out we can be more granular and we decide to offer it; it might constitute a breaking change depending on preferred implementation of a more detailed filter.

You're absolutely right, but the only way to get those study spaces would be to use OAuth tokens to access the data, and we'd also have to maintain which study spaces are available to whom.

> Otherwise, I'm happy with the code itself.
Yay!

PR #1141, opened 2019-03-14
Fixed throttling bug that occurs at midnight

Feverfew:
## What does this PR do?
Fixes throttling bug that occurs at midnight as Redis keys can't have TTL of 0

## ✅ Pull Request checklist

- [ ] Is this code complete?

No

## :squirrel: Deploy notes
None

### DIFF
--- a/backend/uclapi/common/decorators.py
+++ b/backend/uclapi/common/decorators.py
@@ -108,7 +108,7 @@ def throttle_api_call(token, token_type):
     if count_data is None:
         # Conditional fixes bug where a call is made exactly at midnight.
         # Redis cannot have key with TTL of 0
-        r.set(cache_key, 1, secs if secs != 0 else 1)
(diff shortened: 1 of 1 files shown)


backend/uclapi/common/decorators.py
```diff
@@ -106,7 +106,9 @@ def throttle_api_call(token, token_type):
 
     secs = how_many_seconds_until_midnight()
     if count_data is None:
-        r.set(cache_key, 1, secs)
+        # Conditional fixes bug where a call is made exactly at midnight.
+        # Redis cannot have key with TTL of 0
+        r.set(cache_key, 1, secs if secs != 0 else 1)
```
ChristopherHammond13:
```suggestion
        r.set(cache_key, 1, secs if secs > 0 else 86400)
```

Aren't we better off making this the number of seconds until midnight in the following day? Since the clock will have rolled over into the next day technically (i.e. the time is 00:00:00)

Issue #1037, opened 2019-02-13
[Workspaces] Return Sensor Hardware IDs in SVG XML


ChristopherHammond13:
At the moment the `/workspaces/images/maps/live` endpoint can return a PNG map with each sensor position laid over the top using SVG. This works well, but we should also add the hardware ID of each sensor to this XML so that developers can map data returned from `/workspaces/sensors` over the top, allowing interfaces that include, for example, pop-ups when each study space is tapped.

Here is a REALLY bad mockup of how this could look if the data was made available:
![image](https://user-images.githubusercontent.com/295396/52746087-a3db2400-2fd8-11e9-8412-d7bea272c1c8.png)

PR #1145, opened 2019-03-15
Adding sensor ID to live map

Feverfew:
## What does this PR do?
Implements #1037 
Each <g> element in the SVG had an 'id' attribute which corresponds to the sensors id. Additional data on sensors can be retried from our Get Sensors (/workspaces/sensors) endpoint.
## ✅ Pull Request checklist

- [ ] Is this code complete?

No

## :squirrel: Deploy notes
Rebuild docs.


Feverfew:
> Great stuff! Thanks so much for picking this issue up sparkles
> Have you tested it working? I realise that we don't have a proper test suite for this API so I'm trusting you on this, and then we can test in staging

The IDs show up on the correct elements to me. I believe I am correct that a user can parse the XML to get the ID attribute from the <g> element?

PR #1146, opened 2019-03-15
Room capacity fix

Feverfew:
## What does this PR do?
Fixes bug in roombookings/rooms endpoint when capacity is not of type float.
Adds tests to confirm that invalid input is handled correctly

## ✅ Pull Request checklist

- [ ] Is this code complete?

No

## :squirrel: Deploy notes
None


ChristopherHammond13:
@Feverfew Does room capacity have to be a float?

Feverfew:
> @Feverfew Does room capacity have to be a float?

For some reason it is a float in the DB

> @Feverfew Does room capacity have to be a float?

The field is a float itself but we wouldn't get a type error later if I changed it to be an int. However, if someone were to input 55.5, how would you like the API to handle this?

ChristopherHammond13:
@Feverfew I think we should just convert the float to an int and work that way? Working in floats _feels_ wrong given that you can't have half people :P @zipy124 what do you think, Mr CTO?

zipy124:
@Feverfew @ChristopherHammond13 I definitely agree with chris here. Since we're working with a discrete variable we should just cast to int. Float feels very wrong.

PR #1183, opened 2019-03-26
Fixing bug when event has null for lecturerid

Feverfew:
## What does this PR do?
Checks that variable is a string before calling the strip() method.

## ✅ Pull Request checklist

- [ ] Is this code complete?

No

### DIFF
--- a/backend/uclapi/timetable/app_helpers.py
+++ b/backend/uclapi/timetable/app_helpers.py
@@ -287,7 +287,7 @@ def _get_timetable_events(full_modules):
 
                     # Check if the module timetable event's Lecturer ID
                     # exists. If not, we use the Lecturer ID associated
-                    # with the module as a whole. If neither exist then 
+                    # with the module as a whole. If neither exist then
                     # we say that we don't know. It's an ugly hack, but
                     # it works around not all timetabled lectures having
                     # the Lecturer ID field filled as they should.
@@ -335,7 +335,7 @@ def _get_timetable_events(full_modules):
 
                     # Check if the module timetable event's Lecturer ID
                     # exists. If not, we use the Lecturer ID associated
-                    # with the module as a whole. If neither exist then 
+                    # with the module as a whole. If neither exist then
                     # we say that we don't know. It's an ugly hack, but
                     # it works around not all timetabled lectures having
                     # the Lecturer ID field filled as they should.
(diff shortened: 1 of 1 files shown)


backend/uclapi/timetable/app_helpers.py
```diff
@@ -287,18 +287,20 @@ def _get_timetable_events(full_modules):
...
-                            _get_lecturer_details(event.lecturerid)
+                            _get_lecturer_details(event.lecturerid.strip())
+                    elif module.lecturerid:
+                        event_data["module"]["lecturer"] = \
+                            _get_lecturer_details(module.lecturerid.strip())
                     else:
+                        # This will give us 'Unknown' in all fields
                         event_data["module"]["lecturer"] = \
-                            _get_lecturer_details(module.lecturerid)
+                            _get_lecturer_details(None)
```
ChristopherHammond13: Isnt't it just easier to mark this as `None` instead of running it through `_get_lecturer_details`?
```suggestion
```

backend/uclapi/timetable/app_helpers.py
```diff
@@ -287,18 +287,20 @@ def _get_timetable_events(full_modules):
...
+                    if event.lecturerid:
                         event_data["module"]["lecturer"] = \
-                            _get_lecturer_details(event.lecturerid)
+                            _get_lecturer_details(event.lecturerid.strip())
+                    elif module.lecturerid:
+                        event_data["module"]["lecturer"] = \
+                            _get_lecturer_details(module.lecturerid.strip())
                     else:
+                        # This will give us 'Unknown' in all fields
                         event_data["module"]["lecturer"] = \
```
ChristopherHammond13:
```suggestion
                        event_data["module"]["lecturer"] = None
```

backend/uclapi/timetable/app_helpers.py
Feverfew: We need this dictionary: https://github.com/uclapi/uclapi/blob/cd0d327d4f0b7ae189458564bf186005408c66ee/backend/uclapi/timetable/app_helpers.py#L107-L112

However everytime this is None we'll always be doing a needless query, should I refactor _get_lecturer_details such that when lecturer_upi is null we just return the 'Unknown' dict without hitting the DB?

PR #1209, opened 2019-04-04
Unifies the way locking works

zipy124:
## What does this PR do?
previously it appears lock.a meant use bucket a for timetable but lock.a mean do not use a for bookings. We did not notice this when unfying locks.

## ✅ Pull Request checklist

- [ ] Is this code complete?

Yes/No

## :squirrel: Deploy notes
Run two caches to fix buckets as a safe measure


zipy124:
Specifically we should test webhooks since i inverted the locks here. I want to make sure it still works!

PR #1218, opened 2019-04-06
Deletes webhooks

### LINKED ISSUE #1214: Webhooks are not deleted upon App deleting
Create an app, link webhook. delete app. The webhook will still fire.

zipy124:
## What does this PR do?
should fix #1214 so that when we delete an app we also delete its webhook.

## ✅ Pull Request checklist

- [ ] Is this code complete?

Yes/No

## :squirrel: Deploy notes
We need to go through all apps which are deleted and dis-able their webhook.

### DIFF
--- a/backend/uclapi/dashboard/api_applications.py
+++ b/backend/uclapi/dashboard/api_applications.py
@@ -202,6 +202,13 @@ def delete_app(request):
     else:
         app = apps[0]
         app.deleted = True
+        webhook = app.webhook
+        webhook.url = ""
+        webhook.siteid = ""
+        webhook.roomid = ""
+        webhook.contact = ""
+        webhook.enabled = False
+        webhook.save()
         app.save()
 
         keen_add_event.delay("App deleted", {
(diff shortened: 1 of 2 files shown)


zipy124:
Last thing to do, is in the webhooks trigger we make sure if the app is deleted do not fire the webhook, just in-case any slip through the cracks of the deployment method i specified above

backend/uclapi/dashboard/api_applications.py
```diff
@@ -202,6 +202,13 @@ def delete_app(request):
     else:
         app = apps[0]
         app.deleted = True
+        webhook = app.webhook
+        webhook.url = ""
```
ChristopherHammond13: Since `url`, `siteid`, `roomid` and `contact` are all nullable, can we make these `= None` instead? I think it'll be neater.

```suggestion
        webhook.url = None
```

zipy124: @ChristopherHammond13 url is not nullable, everything else can be made none though yes.

PR #1313, opened 2019-05-23
Provides locations for workspaces surveys

### LINKED ISSUE #1165: Provide Coordinates of Each Workspaces Survey
This would help apps like UCL Assistant sort libraries by proximity to user, and navigate users to study space buildings.

zipy124:
## What does this PR do?
as the title says. At cache we insert into the redis some hard coded locations. For code-review we should check all the address' are in the format we want (i wasn't sure about this) and think about potentially a better way to do this. Maybe we can search google maps api with the title of the room?

## ✅ Pull Request checklist

- [x] Is this code complete?

Yes/No

## :squirrel: Deploy notes

## Anything else
closes #1165 
Also fixes existing documentation for surveys endpoint as it was missing data, in the wrong order and also booleans aren't true or false in json but a 1 or 0.


zipy124:
@Feverfew thanks for the link regarding address' formatting, I'll try comb through it all tonight. Just looked a the build fail, i forgot to modify the tests my bad, I'll fix them and try to make sure they also test this new functionality. The docs will be updated at the same time!

zipy124:
Update: Everything is done now apart from normalizing address format

frontend/src/components/documentation/Routes/Workspaces/GetSurveys.jsx
```diff
@@ -121,15 +144,35 @@ export default class WorkspacesGetSurveys extends React.Component {
                             example="false"
                             description="Whether the survey represents a staff workspace (`true`) or a student work or study space (`false`). This is useful in apps where you set the `survey_filter` parameter to `all` and wish to do filtering of student and staff workspaces in your app or API as opposed to leaving the filtering to UCL API. By default, as `survey_filter` is set to `student` unless you specify otherwise, all workspaces will have this parameter set to `false`." />
                         <Cell
-                            name="name"
+                            name="location"
+                            extra="dictionary"
```
Feverfew: "dictionary" is inconsistent with the rest of the docs. "object" is preferred (and currently used), simply because we are sending JSON, not Python data structures.

frontend/src/components/documentation/Routes/Workspaces/GetSurveys.jsx
```diff
@@ -121,15 +144,35 @@ export default class WorkspacesGetSurveys extends React.Component {
                             example="false"
                             description="Whether the survey represents a staff workspace (`true`) or a student work or study space (`false`). This is useful in apps where you set the `survey_filter` parameter to `all` and wish to do filtering of student and staff workspaces in your app or API as opposed to leaving the filtering to UCL API. By default, as `survey_filter` is set to `student` unless you specify otherwise, all workspaces will have this parameter set to `false`." />
                         <Cell
-                            name="name"
+                            name="location"
+                            extra="dictionary"
+                            example='{"coordinates": {"lat": "51.522897","lng": "-0.127864"},"address": ["Newsam Library and Archives","20 Bedford Way","London","WC1H  0AL"]'
+                            description="Dictionary of survey location information including co-ordinates and the address." />
+                        <Cell
+                            name='location["coordinates"]'
```
Feverfew: This (and not only this) name is also inconsistent with the rest of the docs. 
This endpoint should be good enough for you to understand how endpoint responses are usually represented:
https://uclapi.com/docs#timetable/bymodule

zipy124:
> So I've looked at your normalisation attempt, and a thought has just come to me. Not all the address lines have a length of 4.

The list length of 4 is simply what we were using in our /rooms endpoint so i assumed it would make sense to keep it as a standard format?

> I assume you've tested both the lat/long and addresses in Google Maps. If you've noticed that maybe sometimes the address doesn't take you to exactly where its located and the long/lat is sometimes more accurate then please mention so in the docs.

The latitude and longitude are always the most accurate location data, hence why even the ucl library website uses these when you click on the "maps" link on a library page provided by UCL. I'll stick a note at the top of the docs to make sure users are aware of this, considering often the address is technically just UCL as each individual building is technically inside the "UCL campus" address.

> Although, in note in determining the recommendation (if any), note that an entrance to the building is probably more helpful than the exact location of a room (seeming as we don't have indoor navigation yet cry @hueyy).

The problem with this is that a lot of buildings have two entrances, or more for example the wilkins building can be entered from south, east or west entrances. It also allows you to differentiate between rooms in the same building, E.g the senit suite and the main library. Although you also have the height difference that some are on the 5th floor, some on the 1st floor etc...

frontend/src/components/documentation/Routes/Workspaces/GetSurveys.jsx
zipy124: ok, I'll replace that, i forgot json calls it an object and not say dictionary, or map.

frontend/src/components/documentation/Routes/Workspaces/GetSurveys.jsx
zipy124: The whole docs are inconsistent, for example in that link we do not show the attributes of the Lecturer Object, but we do for other Objects? If you mean the quotes they will be removed i didn't notice we don't include those (even though it makes more sense to me).

backend/uclapi/workspaces/occupeye/cache.py
```diff
@@ -118,7 +118,7 @@ def cache_survey_data(self):
             staff_survey = str(
                 int(survey["SurveyID"]) in self._const.STAFF_SURVEY_IDS
             )
-
+            location_data = self._const.SURVEY_LOCATIONS[survey["Name"]]
```
ChristopherHammond13: What if it doesn't exist? They add surveys without telling us.

zipy124: Added a null condition!

hueyy:
Updated the docs.

Issue #963, opened 2019-01-27
Compress Frontend Assets


ChristopherHammond13:
- JPG all the things
- gzip all the things
- Make sure S3 is happy to serve gzipped content

zipy124:
In the meantime images have been compressed as small as png will allow. not sure JPG is required tbh. gzip would be good though. ref: #1058

hueyy:
Might I suggest we use [Cloudinary](https://cloudinary.com) for compressing and serving images? That way we don't need to maintain and tweak an image optimisation solution, and we can easily serve assets based on the capabilities of the browser (e.g. WebP to Chrome and JPEG-XR to IE).

There's a free tier (25GB of storage) which seems more than sufficient for our needs.

hueyy:
Resolved in #1347

Issue #962, opened 2019-01-27
ES6 Migration and Tree Shaking


ChristopherHammond13:
Huey mentioned that the next stage to optimise our frontend is to strip out unused dependencies and fully migrate to ES6. This would be awesome, but it's not a high priority at the moment.

I've assigned @hueyy and @anushks as they are the most experienced with frontend, but anyone can jump on this if they have interest.

hueyy:
Resolved in #1347

PR #1347, opened 2019-06-02
Compress frontend assets

hueyy:
## What does this PR do?

Addresses #962 and #963 

* ES module tree shaking for dependencies that support it
* More selective imports
* Replaced momentjs with much smaller dayjs
* Removed `prop-types` in production

Decided not to gzip at the build stage in the end because it complicates the build process. Django gets confused when it has to serve gzipped assets. Anyway, Cloudflare gzips our static assets already.

I think Snyk is upset about v0.2 of `material-ui` which has been deprecated and contains a number of vulnerabilities. We've partially migrated to v4.0, but there are many parts that still rely on the old version. We can ignore Snyk for now. I'll send in another PR removing v0.2 of `material-ui` later.

**Probably best to merge & squash**

## ✅ Pull Request checklist

- [x] Is this code complete?

No

## :squirrel: Deploy notes
Need to build & redeploy static assets


ChristopherHammond13:
So glad we're doing this! I think we should have assets gzipped automatically by CloudFlare but it'll be great to have the smaller assets in S3 to start with so that we don't have to wait for caching.

hueyy:
@zipy124 merge conflicts resolved!

ChristopherHammond13:
This is super exciting! Great work @hueyy

PR #1350, opened 2019-06-03
About page merge

TuxedoFish:
## What does this PR do?
Adds in Anirudh's about page and makes it look even nicer and includes the past year. Had to put it on a new branch as I just found there were a lot of things in the branch that were way behind master.

## ✅ Pull Request checklist

- [ ] Is this code complete?

Yes/No


zipy124:
After looking at the screenshot above, would it not make sense to have the years the other way around? like start with the current team then delve into history?

TuxedoFish:
Will remake design using content created once refactor is merged in

PR #1541, opened 2019-08-19
Update medium articles for use on the frontend

TuxedoFish:
## What does this PR do?
Adds extra information to the current gather of the medium articles so that the information can be used on the front end to make it look a little prettier 😜. Should just add on to what was already there.

## ✅ Pull Request checklist

- [ ] Is this code complete?

Yes/No

## :squirrel: Deploy notes


TuxedoFish:
Just saw the test case will update tonight!

PR #1693, opened 2019-10-09
Changed hound to enable scss comments and to ignore migrations

TuxedoFish:
## What does this PR do?
Changes the hound settings for python migrations. Attempts to resolve the scss hound comments so we can use that for the future!


TuxedoFish:
Using the PR to test whether settings are correctly set up

Issue #1566, opened 2019-08-24
Embed a linter in pre-commit to sate Hound


hueyy:
A formattor/linter like [autopep8](autopep8) or [black](https://github.com/psf/black) can automatically check for pep8 violations and fix them (where possible)

We could set it up with a pre-commit hook so it automatically runs on every commit.

hueyy:
* Add all our preferred linting rules in .eslintrc
* Write a pre-commit hook that runs eslint

PR #1753, opened 2019-10-27
Add pre-commit hook

### LINKED ISSUE #1692: HoundCI Configuration
Need to sort out hound issues with current scss! Also try to add in configuration files so hound is more useful, only pointing out issues that are resolvable / our code.

hueyy:
## What does this PR do?
Adds a pre-commit hook that does the following:

* eslint with autofix for JS
* stylelint with autofix for scss
* autopep8 followed by flake8 for python code. It runs both because only autopep8 will automatically fix formatting issues, and only flake8 will catch errors.

The linting will only run on staged files.

## ✅ Pull Request checklist

- [x] Is this code complete?

No

## :squirrel: Deploy notes
All devs will have to update their node dependencies (i.e. run `npm i`) before the pre-commit hook is added to their local dev environment.

## Anything else

Hound doesn't support a number of eslint plugins. These have to be whitelisted by the Hound team, which doesn't seem very responsive. I've therefore created a separate `.eslintrc.hound.json` config with fewer plugins just for Hound.

Resolves #1692 and #1566.


TuxedoFish:
Super helpful resolving these sorts of issues before you push! Resolved the issues in the rest of the pages so in future commits I dont have a million things! Happy for it to be merged, just need to be careful and not force quit the linting can cause unexpected behaviour

TuxedoFish:
Update: just noticed hound complains about resolved paths but the linting didnt

hueyy:
I'm not sure why Hound isn't satisfied. It was perfectly happy earlier on. The output from Hound looks like an error message, so I'm guessing it could be a bug in Hound:
>Resolve error: package.json not found in path import/default
Resolve error: package.json not found in path import/namespace
Resolve error: package.json not found in path import/no-extraneous-dependencies

TuxedoFish:
> I'm not sure why Hound isn't satisfied. It was perfectly happy earlier on. The output from Hound looks like an error message:
> 
> > Resolve error: package.json not found in path import/default
> > Resolve error: package.json not found in path import/namespace
> > Resolve error: package.json not found in path import/no-extraneous-dependencies
> 
> It runs just fine, and there is no error locally, so I think this is a Hound bug of some sort.
> If Zak and Alex agree, I think this can be merged.

Yeah I didn't understand this either? It happened after I just got the other files to be included in pages/

PR #1892, opened 2019-11-30
Update locations of workspaces

zipy124:
## What does this PR do?
Updates locations for workspaces

## ✅ Pull Request checklist

- [ ] Is this code complete?

Yes/No

## :squirrel: Deploy notes


zipy124:
> The fact that we have to do this now all the time annoys me
Yeh, It would help if there weren't arbitrary name changes by UCL.

Issue #1757, opened 2019-10-28
[BUG] Dates on the dashboard are incorrect


hueyy:
**Describe the bug**
The created at and last modified dates for apps on the dashboard are incorrect.

![image](https://user-images.githubusercontent.com/6523121/67673142-48d43c00-f971-11e9-89fa-711589d1cca4.png)

(this app was not created on 28th October; it was created several months before)

hueyy:
![image](https://user-images.githubusercontent.com/6523121/69173182-29a37700-0af7-11ea-9949-765bf6c93e31.png)

PR #2070, opened 2020-02-07
fix oauth myapps

TuxedoFish:
## What does this PR do?
Fixes the app settings page and introduces instructions on how to test oauth apps quickly:

Testing the settings page (locally)

1: Ensure running through ./manage.py runserver NOT docker:

a)	Start services through sudo service – start for postgres and redis (postgresql, redis-server)
b)	Activate virtual environment
c)	Start server ./manage.py runserver
d)	Steps b and c for fakeshibboleth and also uclapi repos (uclapi – 8000, fakeshibboleth – 8001)

2. Generate a token to play around with:

a)	Pull down the django-uclapi-oauth repository from github
b)	Run the repository on port 8002 [ No need for virtual environment ]
c)	http://localhost:8000/oauth/token?client_id=INSERT_APP_ID_HERE&state=1 
d)	Go to uclapi (db) > dashboard_app (table) and manually change callback to, http://localhost:8002/uclapi/callback so that you can use the url in step c)
e)	Go into the .env of the repo and change the variables to reflect your app
f)	Go to the link in step c) and follow through the authorisation and you will have approved a scope
g)	Bear in mind when the login page pops up remember to change the url to auto/login from /login and refresh

3: Manipulate token:

a)	Navigate to uclapi (db) > oauth_oauthtoken (table) 
b)	Should have a token for your app now which you can enable and disabled manually to test the settings page
c)	You can now also click deauthorise on the settings page to deauthorise the app which deletes the token so you will have to repeat bits of 2.

## ✅ Pull Request checklist

- [ ] Is this code complete?

Yes/No

## :squirrel: Deploy notes
Just aesthetic

## Anything else

settings-uclapi


hueyy:
Looks great!

image


A few things that I went ahead and helped resolve:

1. padding issues
2. the date bug (#1757) that keeps popping up for some reason. I'm not sure why we keep regressing on this. Probably have to be more careful when working on old branches and updating with master.

A few other things:

1. I changed the phrasing of the confirmation message. It was fine before, but I think this is slightly clearer.
1. Step 2(c) on your instructions above is a typo I think. It should be /oauth/authorise.
2. There should be a logout button somewhere, probably in Settings.
3. Dashboard and Settings can be confusing => what's the difference? Perhaps at some point we should combine the two into a single Settings page with a Developer tab and a User tab.
4. It looks like there are 2 duplicate files: `loginlayout.jsx` and `LoginLayout.jsx` which I suspect is because you renamed the file and Windows ignores capitalisation but github doesn't.

TuxedoFish:
> Looks great!
> 
> image
> 
> A few things that I went ahead and helped resolve:
> 
> 1. padding issues
> 2. the date bug (#1757) that keeps popping up for some reason. I'm not sure why we keep regressing on this. Probably have to be more careful when working on old branches and updating with master.
> 
> A few other things:
> 
> 1. I changed the phrasing of the confirmation message. It was fine before, but I think this is slightly clearer.
> 2. Step 2(c) on your instructions above is a typo I think. It should be /oauth/authorise.
> 3. There should be a logout button somewhere, probably in Settings.
> 4. Dashboard and Settings can be confusing => what's the difference? Perhaps at some point we should combine the two into a single Settings page with a Developer tab and a User tab.
> 5. It looks like there are 2 duplicate files: `loginlayout.jsx` and `LoginLayout.jsx` which I suspect is because you renamed the file and Windows ignores capitalisation but github doesn't.

Cheers Huey 💃 

Changed the instructions, sorry about it being such a pain to debug!! Agree with putting dashboard and settings into one as 2 different tabs in the future. I will add a log out button to this and delete that duplicate file. Glad that it is working for you too!

Issue #1174, opened 2019-03-22
Dashboard Redesign


hueyy:
Revamp the functional but unaesthetic dashboard to something resembling @anushks's concept art.

[![concept art](https://user-images.githubusercontent.com/6523121/54850260-96ebd780-4cde-11e9-969d-a49eae6374f1.png)](https://www.figma.com/file/NWyEyi81HEnrU70d7KsL5r3R/ConceptDashboardNew?node-id=0%3A1)

PR #2102, opened 2020-02-19
Feature/dashboard redesign

TuxedoFish:
## What does this PR do?
#1174  Makes the dashboard more consistent with the rest of the website and hopefully easier to use. Using Anush's design from a while ago! Ill leave a bunch of screenshots. In terms of functionality, all of the "actions" or updates to the backend are defined within the Dashboard page. The App component has all of the ui in it and just makes calls to these actions to update the state and persistance of the app. Added a few new components like  Field, CheckBox and Icon which can be used in the future hopefully!

## ✅ Pull Request checklist

- [ ] Is this code complete?

Yes/No

## :squirrel: Deploy notes

## Anything else
full-options
Main
app


hueyy:
* Two kinds of fields: one with the save button and one without
* Make entire projects collapsible
* UCL Market => Marketplace
* Allow user to start editing field immediately upon clicking text in field
* Include link to documentation about oauth in the dashboard
* Submit button on new project modal
* Padding on the project divs
* New project button at the top

TuxedoFish:
@hueyy @anushks Have revamped the initial pull request and attached photos at the top and then made a few changes. Let me know if you have any more suggestions for things to do. I will try sort hound out soon!

hueyy:
Nice work on the padding. That looks perfect now. 

Some other points:

It's not super intuitive what the scope buttons do if you aren't familiar with the previous dashboard. Might be best to stick to checkboxes.

image

I like the fact we can now distinguish between read-only and editable fields based on colour, but I do worry a bit about accessibility. The shades of blue are pretty close and I wouldn't be surprised if some people cannot distinguish between the two. 

image

I find the arrows/chevrons very odd-looking, especially the vertical ones, probably because they're just rotated Vs. Try using the chevrons from the feather icon pack or something? I also think it might look better if the app names and chevron were left-aligned rather than centre-aligned.

image

There's no feedback when the copy button is pressed so I wasn't sure if it worked. This is a minor issue, but I think the experience could be improved if there was a small feedback message (something like a toast).

This PR is getting pretty large, so one possibility is to merge this in first, and tackle these other issues in a subsequent PR.

TuxedoFish:
> 
> 
> Nice work on the padding. That looks perfect now.
> 
> Some other points:
> 
> It's not super intuitive what the scope buttons do if you aren't familiar with the previous dashboard. Might be best to stick to checkboxes.
> 
> image
> 
> I like the fact we can now distinguish between read-only and editable fields based on colour, but I do worry a bit about accessibility. The shades of blue are pretty close and I wouldn't be surprised if some people cannot distinguish between the two.
> 
> image
> 
> I find the arrows/chevrons very odd-looking, especially the vertical ones, probably because they're just rotated Vs. Try using the chevrons from the feather icon pack or something? I also think it might look better if the app names and chevron were left-aligned rather than centre-aligned.
> 
> image
> 
> There's no feedback when the copy button is pressed so I wasn't sure if it worked. This is a minor issue, but I think the experience could be improved if there was a small feedback message (something like a toast).
> 
> This PR is getting pretty large, so one possibility is to merge this in first, and tackle these other issues in a subsequent PR.
> 
> On a sidenote, I haven't really customised stylelint for our purposes, so if you think any rules are annoying and unnecessary, feel free to disable them.

I agree with the points on the *arrows, *scopes, and adding feedback to the copy button. These wont be too hard to change. I will have a go at that tonight!

The colors I also agree with but I feel like that might be something we need to discuss more in depth so maybe in a future pr.

Sorry taking ages to reply, still havent got husky to run pre commit so I decided to just manually run it now! 

Thanks for reviewing it so quickly

hueyy:
These Hound errors:

> Resolve error: package.json not found in path import/default
Resolve error: package.json not found in path import/namespace
Resolve error: package.json not found in path import/no-extraneous-dependencies

appear to be a reoccurrence of https://github.com/houndci/hound/issues/1637.

I'm ignoring them and merging this in because linting runs fine locally.

Issue #2076, opened 2020-02-08
[Feature Request] Text, not just icon should be clickable


hueyy:
![image](https://user-images.githubusercontent.com/6523121/74085148-ec071800-4a6d-11ea-984d-306288cdf82e.png)

Right now, only the icon is clickable. Clicking the text should also bring the user back to the homepage.

hueyy:
Resolved by #2114

Issue #2075, opened 2020-02-08
[Feature Request] Remove header on Marketplace


hueyy:
On the app page in Marketplace, the top header may not really be necessary. It's great on the main marketplace page (`/marketplace`), but increases scroll distance on app detail pages.

![image](https://user-images.githubusercontent.com/6523121/74085031-ee1ca700-4a6c-11ea-995e-bbb056513ba5.png)

hueyy:
Resolved by #2114

Issue #2074, opened 2020-02-08
[Feature Request] Move install buttons on Marketplace to top


hueyy:
The current app view in the Marketplace is very similar to the standard app store layout. I think users might expect the `GET` or `INSTALL` button to be in the same place.

Might make sense to move the buttons to be next to the title.

![image](https://user-images.githubusercontent.com/6523121/74085001-7f3f4e00-4a6c-11ea-9e8c-ec55de752d23.png)

hueyy:
Resolved by #2114

PR #2114, opened 2020-02-23
Feature/aesthetic changes

TuxedoFish:
## What does this PR do?
Addresses Issues: #2074 #2075 #2076 brought up by @hueyy .  Thanks for bringing these up, has made it look a lot better doing those things!

Also the screenshots were just so ugly and I couldn't think of a good way to make it look good without bringing in a dependency so the screenshots now use the react-responsive-carousel with a couple of images in the background. Looks a lot more slick now with that change so I think this is a dependency thats worth it and is only used for one thing

## ✅ Pull Request checklist

- [ ] Is this code complete?

Yes/No

## :squirrel: Deploy notes

## Anything else
screencapture-localhost-8000-marketplace-uclcssa-2020-02-23-08_07_02

screencapture-localhost-8000-marketplace-uclcssa-2020-02-23-08_05_32


TuxedoFish:
@hueyy Fixed the social media and things from Sunday and it seems to be working fine now!

PR #2120, opened 2020-02-24
Remove material-ui v0.2

hueyy:
## What does this PR do?
This removes the v0.2 deprecated version of `material-ui`. The current version (already installed) is v4.0.

## ✅ Pull Request checklist

- [x] Is this code complete?

No

## :squirrel: Deploy notes


TuxedoFish:
This is such a good thing to do, genuinely couldn't tell the difference between the old and new side bar. Couple of comments, on the social medias facebook has a lower case f should be Facebook to fit in with the others. Also when you click drop downs and the scroll bar appears sometimes twitter will go off the edge of the screen and be unaccessible. (These are super minor points and im not even sure if we need the social media on the side bar now that they are all at the bottom of the screen). The off screen issue only happens on the desktop side bar.

No observable differences otherwise, on a side note the material ui components in SettingsView should be replaced with ButtonViews instead of importing material ui there but can do that in a future pr. + all of the linting stuff makes the code so much more readable.

hueyy:
@TuxedoFish thanks for the quick review. I can't seem to find the lowercase `f`. Is it on the sidebar or somewhere else? Also can't seem to replicate the Twitter going off screen bug. That's a bug on the current prod version, but I think I fixed that when I removed `material-ui v0.2`.

TuxedoFish:
Take 2: on the right branch now 🥇. 

Just taking a look at the side bar it looks way more smooth on desktop than before. When I try to use the side bar on mobile devices, I get a react error in the console. 

Failed prop type: Invalid prop `children` supplied to `ForwardRef(SwipeableDrawer)`, expected a ReactNode.

Also appears that some of the documentation containers are not full width 

image

hueyy:
@TuxedoFish good catch. Fixed both issues.

TuxedoFish:
One other thing I just noticed is in the documentation the parameters seem to be going onto 2 lines as opposed to one which gives it quite a large padding:

image

hueyy:
@TuxedoFish I've tested it on a number of screen sizes, but I can't seem to reproduce that spillover:

image

Maybe check if your branch is up to date?

TuxedoFish:
@hueyy yeah my bad just checked again! Will merge

Issue #2117, opened 2020-02-24
[BUG] Hamburger menu does not work on mobile


hueyy:
**Describe the bug**
Pressing the 3-bar hamburger button in the header does nothing on mobile.

**Expected behavior**
I would have expected a menu to open up.

Issue #970, opened 2019-01-29
There is no logout button on the dashboard?!?!


zipy124:
how has no one mentioned this before?

zipy124:
so there's no way of switching between ucl accounts on the dashboard? since that would require logging out of shibboleth?

ChristopherHammond13:
@zipy124 correct. The way to do it right now is to delete local session tokens for both us *and* UCL, and then log in again. Even if you go to `https://uclapi.com/Shibboleth.sso/Logout`, it will only kill the session token for Shibboleth on our servers (assuming you haven't switched servers via the load balancer as the sticky window is only 5 minutes until Dockerisation is rolled out). That means clicking Dashboard will still go:
- `/dashboard`/
- `/Shibboleth.sso/Login?target=/dashboard/shibcallback`
- UCL Sign in
- UCL's session will automatically log you in
- `/dashboard/shibcallback`
- `/dashboard` with the same account as before

(uclapi.com omitted from above URLs for brevity)

ChristopherHammond13:
Can't fix won't fix will never fix sorry not sorry?

Feverfew:
When this issue was opened, we did not have a logout button at all. Now we have a logout button. As briefly discussed in #2170, I now believe that neither solution is satisfactory from a security point of view.

Having no logout button at all does not allow a user to logout from their session (obviously).

Having a logout button gives a false sense of security, as one can log back in, without credentials, as Shibboleth would still have your session.

A sensible compromise would be to allow logout, but to then make it clear that further action is needed to achieve a _proper_ logout (i.e., close your browser window). @TuxedoFish came up with a simple javascript alert, something nicer should be aimed for. I won't comment on aesthetics, but whatever conveys the point clearly, isn't terrifyingly ugly and can't be missed, I'll approve

TuxedoFish:
@Feverfew I agree I wasn't actually aware of the fact that Shibboleth would still have your session but now that I do, it seems like the best of the options to have a logout button but to alert the user that they need to take extra steps.

The javascript alert was definitely not the best solution, I thought it could work temporarily until we make a new page but I get that this seems a bit ugly!! I will have a go at creating a page that has a warning like this and tag it here once I open it!

PR #2170, opened 2020-03-11
Fix/settings page flow

### LINKED ISSUE #2110: Broken logging in flow for the settings page
**Describe the bug**
Logging in when accessing settings page leads to 404

**To Reproduce**
Clear your cache 
Go to uclapi.com
Click on settings 
Log into your account
Observe 404 uclapi.com/shibcallback does not exist:

**Expected behavior**
Should redirect to the callback for the settings page

OR 

Could redesign so the log in button redirects to the home page and then once logged in those pages are accessible

**Screenshots**
image

TuxedoFish:
## What does this PR do?
Fixes #2110 so that you can log in by clicking on the settings page. #2117 Has been resolved in an older commit lost where it got merged in.

## ✅ Pull Request checklist

- [ ] Is this code complete?

Yes/No

## :squirrel: Deploy notes


backend/uclapi/uclapi/urls.py
```diff
@@ -26,6 +27,8 @@
     url(r'^dashboard/', include('dashboard.urls')),
     url(r'^docs', documentation),
     url(r'^about', about),
+    url(r'^settings', settings),
+    url(r'logout$', logout),
     url(r'^marketplace', marketplace),
```
Feverfew: Ok, looking at the diff I've noticed I've noticed an issue.

First of all,  all these regexes I've commented on are wrong. Note that in Python regexes have the following format defined [here](https://docs.python.org/3/library/re.html). In particular you'll notice that `^` matches the start of the input and `$` matches the end of the input. Currently you'll be taken to the about page if you navigate to `about`, correctly, but also to `about123`, incorrectly. The inverse happens with `logout`, I will be taken the logout page if I navigate to `i-dont-want-to-logout`, ironically. The solution? All these regexes highlighted above require a `^` at the beginning and a `$` at the end. 

HOWEVER, only correct for the lines changed in this current diff. For the other lines that aren't changed please fix them in a separate PR. I would like to note that I believe `url()` is deprecated since Django 2.0 (which thanks to Zak, we've finally switched to), and `path()` is recommended instead. You can find a way to switch to `path()` [here](https://consideratecode.com/2018/05/02/django-2-0-url-to-path-cheatsheet/). I'll accept a diff that either uses `url()` correctly or switches to `path()` (also correctly), but `path() would be prefferred.

frontend/src/components/layout/NavBar.jsx
```diff
@@ -18,7 +18,7 @@ import Link from './Link.jsx'
 const links = [
   {
     name: `settings`,
-    link: `/oauth/myapps`,
```
Feverfew: @hueyy @anushks, noting my comment above about the incorrect regex in our Python code, is this doing what we expect from a React PoV?

backend/uclapi/oauth/urls.py
```diff
@@ -4,7 +4,6 @@
 
 urlpatterns = [
     url(r'myapps/shibcallback$', views.myapps_shibboleth_callback),
-	url(r'myapps', views.my_apps),
     url(r'authorise/$', views.authorise),
     url(r'shibcallback', views.shibcallback),
```
Feverfew: As a note for a new diff to fix the issue I described earlier, this only requires a `$`, assuming you keep using the `url()` function. If you switch to `path()`, hopefully the document I linked should describe what you should do.

backend/uclapi/oauth/urls.py
```diff
@@ -4,7 +4,6 @@
 
 urlpatterns = [
     url(r'myapps/shibcallback$', views.myapps_shibboleth_callback),
-	url(r'myapps', views.my_apps),
```
Feverfew: Why is this being removed?

backend/uclapi/oauth/views.py
```diff
@@ -763,6 +763,10 @@ def deauthorise_app(request):
 
 @ensure_csrf_cookie
 def logout(request):
```
Feverfew: I've been thinking about the point of logout, especially considering issue #970. The conclusion from the issue, AFAICT is that we shouldn't do logout, as it is misleading from a security PoV.

Considering the security aspect, I think we should remove the logout button, view, and url pattern completely. People will logout, and then may give their laptop to someone else, assuming they can't login, when in fact they can. If we remove the option to logout, then giving your laptop to someone else is a security flaw committed by the user willingly, knowing the dangers.

frontend/src/components/layout/NavBar.jsx
hueyy: These values go into the href attributes of anchor tags. No regex at work here, so looks fine.

backend/uclapi/oauth/views.py
hueyy: Library has the same issue. Their solution is to ask people to close their browser window after they've logged out. This works presumably because shibb cookies are session cookies and expire once the browser session closes.

Perhaps we can keep the logout button but include a warning explaining that users aren't really logged out until they close their browser: e.g.

>You have been logged out of UCL API. However, you are still logged in to other UCL services. To log out completely, please close your web browser.

backend/uclapi/oauth/urls.py
TuxedoFish: Moving the settings page from /oauth/myapps => /settings because it makes more sense logically based on our navigation

backend/uclapi/uclapi/urls.py
TuxedoFish: Okay have switchedd to using Path for urls in this pull request and I will update the other ones in a future pull request too. Thanks for the heads up!

backend/uclapi/oauth/views.py
TuxedoFish: I am happy to do either, I feel like i we add a warning explaining that you need to close your browser it at least gives the user the option to be able to log out and instructions on the website without clearing all of their cookies.

I can add this in later, do you think the warning that they need to close the browser to completely log out is sufficient @Feverfew

TuxedoFish:
Okay, I have changed the url => path for the urls that I have changeed in this pull request. I will change the others in a seperate pull request.

I have added a pop up warning when you log out like this:

image

@Feverfew @hueyy do we feel like this solves the security issues

hueyy:
I'd prefer if it was a separate page, simply because it's too easy to accidentally close an alert box like this, but this works for now.

TuxedoFish:
> I'd prefer if it was a separate page, simply because it's too easy to accidentally close an alert box like this, but this works for now.

Oh okay so kind of like the authorise page and you have to click I understand to go back to the home page?

backend/uclapi/uclapi/urls.py
```diff
@@ -16,8 +16,9 @@
 from django.conf.urls import url, include
 from django.contrib import admin
 from dashboard.views import documentation, home, about
+from oauth.views import settings, logout
 from marketplace.views import marketplace
-
+from django.urls import path
```
Feverfew: Please move this line to just under `from django.contrib import admin`

TuxedoFish: Done!

TuxedoFish:
> Ok much better.
> Please address the following, then I'll approve:
> 
> 1. Address my one new inline comment (that isn't resolved).
> 2. Remove the alert (and keep logout): this should be discussed in #970 and implemented in a different PR.
> 
> In addition, with the other URLs that I've noticed are incorrect, please correct them in a separate PR.

Okay perfect, I have disabled the alert and moved the import statement, going to change the urls=>path

frontend/src/pages/HomePage.jsx
```diff
@@ -37,6 +37,12 @@ class HomePage extends React.Component {
     let loggedIn = false
     if (window.initialData.logged_in === `True`) { loggedIn = true }
 
+    if (window.initialData.has_just_logged_out === `True`) { 
+      // TODO: Decide on a way to let user know they have not been properly logged out
+      //alert("You have been logged out of UCL API. However, you are still logged in to "
+      //  + "other UCL services. To log out completely, please close your web browser.")
+    }
+
```
Feverfew: Not necessary for this PR I believe

TuxedoFish:
@Feverfew Removed all of those ^^

backend/uclapi/dashboard/views.py
```diff
@@ -199,6 +199,7 @@ def home(request):
     except KeyError:
         logged_in = False
 
+
```
Feverfew: Remove extra line

TuxedoFish:
@Feverfew Took out the extra line and did squash and merge

PR #2176, opened 2020-03-12
Fix webpack config

hueyy:
## What does this PR do?

This avoids using MiniCssExtract due this [issue](https://github.com/webpack-contrib/mini-css-extract-plugin/issues/386) and uses StyleLoader instead for production builds.

## ✅ Pull Request checklist

- [x] Is this code complete?

No

## :squirrel: Deploy notes
Rebuild frontend

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -3404,9 +3404,9 @@
       "dev": true
     },
     "acorn": {
-      "version": "7.1.0",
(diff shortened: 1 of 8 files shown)


TuxedoFish:
Seems to fix the problems that we had with the marketplace when webpack is built using the prod file, @hueyy am I okay to merge this in?

Issue #2236, opened 2020-03-31
[BUG]


TuxedoFish:
**Describe the bug**
Trying to navigate to app pages on the marketplace brings up the 404 page becuase the link is defined in react but not in django

**To Reproduce**
1. request endpoint with these parameters
2. Notice a field is null but shouldn't be or un-expected error message etc...

**Expected behavior**

**Screenshots**

**Additional context**

PR #2237, opened 2020-03-31
Added a path to django so marketplace app pages will render

TuxedoFish:
## What does this PR do?
Solves #2236 by adding a path in for the /marketplace/<app_id> urls that react redirects you to when you click on an app icon. Also made Unikomet a featured app because it should be since it is the most recent one!

## ✅ Pull Request checklist

- [ ] Is this code complete?

Yes/No

## :squirrel: Deploy notes


backend/uclapi/marketplace/views.py
```diff
@@ -1,5 +1,5 @@
 from django.shortcuts import render
 
 
-def marketplace(request):
+def marketplace(request, id=None):
```
Feverfew, line 4: This requires a comment, on most linters this will pick up as an unused variable.

TuxedoFish: So we don't actually need to pass the id to react because react router already has it. But django needs to have the path /marketplace/:appid defined but not do anything with the id, when you say requires a comment shall I just explain that over the top or change it?

Feverfew: Yes, the explanation that you've given now needs to be in a comment above the function definition (modulo the grammar errors).

TuxedoFish: Haha! Okay I will add that, thanks!

Issue #2372, opened 2020-05-15
[Feature Request] Do not show authorize screen if already authorized


zipy124:
**Is your feature request related to a problem? Please describe.**
When using UCLAPI Oauth as a login, each time the user logs in the user has to re-authorize the app. It seems redundant to ask the user to re-authorize each time since if already authorized logging in again shows the intent to re-authorize.

**Describe the solution you'd like**
If already authorized, skip the page and simply return the code as if the user had clicked allow. This is now a viable solution since a user can revoke permissions directly in the settings page of our site.

PR #2373, opened 2020-05-15
Oauth auto allow

zipy124:
This addresses issue #2372

### DIFF
--- a/backend/uclapi/oauth/views.py
+++ b/backend/uclapi/oauth/views.py
@@ -229,40 +229,40 @@ def shibcallback(request):
 
     try:
         token = OAuthToken.objects.get(app=app, user=user)
-        if token.scope.scopeIsEqual(app.scope) and token.active:
-            code = generate_random_verification_code()
-
-            r = redis.Redis(host=REDIS_UCLAPI_HOST)
-
-            verification_data = {
-                "client_id": app.client_id,
-                "state": state,
-                "upi": user.employee_id
-            }
-
-            verification_data_str = json.dumps(
-                verification_data, cls=DjangoJSONEncoder)
-
-            # Store this verification data in redis so that it can be obtained
-            # later when the client wants to swap the code for a token.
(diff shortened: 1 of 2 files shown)


backend/uclapi/oauth/views.py
```diff
@@ -228,9 +226,46 @@ def shibcallback(request):
...
+            # The code will only be valid for 90 seconds after which redis will
+            # just drop it and the process will be invalidated.
+            r.set(code, verification_data_str, ex=90)
+
+            # Now redirect the user back to the app, at long last.
+            # Just in case they've tried to be super clever and host multiple
+            # apps with the same callback URL, we'll provide the client ID
+            # along with the state
+            return redirect(
+                app.callback_url + "?result=allowed&code=" + code +
```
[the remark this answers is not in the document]
zipy124: Hound doesn't accept either before or after it seems? great....

backend/uclapi/oauth/views.py
```diff
@@ -228,9 +226,46 @@ def shibcallback(request):
     }
 
     initial_data = json.dumps(page_data, cls=DjangoJSONEncoder)
-    return render(request, 'permissions.html', {
-        'initial_data': initial_data
-    })
+
+    try:
+        token = OAuthToken.objects.get(app=app, user=user)
```
Feverfew, line 231: AFAICT this is the only line that can a `OAuthToken.DoesNotExist` exception, so the try/catch block should only surround this line.

zipy124: That makes a lot of sense, should probably do that in the rest of the file too retroactively at some point.

zipy124:
> 
> 
> LGTM code wise. I'd like to see some unit tests though, I just can't remember off the top of my head if we have infrastructure for automatically testing functions based on shibcallbacks...

Just had a look, it would appear we do have testing for that. I'll look at adding it in now

PR #2376, opened 2020-05-16
Improve test coverage of OAuth + PEP8 compliance

zipy124:
So far this contains:
Tests:
Full test coverage for oauth/user/deny
Partial test coverage for oauth/user/allow
PEP8 Removed bare exception in first try block in /deny
Added TypeError to catch the NullType exception when post data is empty for first try block in allow and deny
Added IndexError to catch the case where user[0] fails at in oauth/user/deny line 311

Testing infrastructure:
Renamed uclapi/celery.py -> uclapi/ourcelery.py to avoid conflicts with the actual celery module


backend/uclapi/oauth/tests.py
```diff
@@ -758,6 +758,121 @@ def test_auto_accept_shibcallback_real_account(self):
...
+        response = self.client.get(
+            '/oauth/user/allow',
+            {}
+        )
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.json()["error"],
+                         ("The signed data received "
+                          "was invalid."
+                          " Please try the login process again."
+                          " If this issue persists, please contact support."))
```
Feverfew, line 771: What support? IMO all error strings should be telling them to contact us (or ISD?).

zipy124: Yes they should probably all say email isd.apiteam@ucl.ac.uk , I will create an issue for that.

zipy124:
> I believe testing status code's by themselves is sufficient... Error messages may change over time (and shouldn't break tests IMO).

This is true, however the error message checking serves the important purpose of checking which part of the code actually breaks and which try block it is in.

Feverfew:
> > I believe testing status code's by themselves is sufficient... Error messages may change over time (and shouldn't break tests IMO).
> 
> This is true, however the error message checking serves the important purpose of checking which part of the code actually breaks and which try block it is in.

Ahh yes that is true. I forgot that we define the error messages in the docs. Although currently the mapping is: small string -> detailed string. It may be better to just have a mapping: error code -> error string. I'll add an issue about that so we can discuss further.

zipy124:
> > > I believe testing status code's by themselves is sufficient... Error messages may change over time (and shouldn't break tests IMO).
> > 
> > 
> > This is true, however the error message checking serves the important purpose of checking which part of the code actually breaks and which try block it is in.
> 
> Ahh yes that is true. I forgot that we define the error messages in the docs. Although currently the mapping is: small string -> detailed string. It may be better to just have a mapping: error code -> error string. Then we can start migrating from comparing strings to comparing error codes (not HTTP error codes btw, but our ones that we define), and then we can change the error string as we see fit (as the message itself doesn't have to be API-compatible) without having to change all the tests as a result...
> 
> I'll add an issue about that so we can discuss further.

Sounds good to me, that was the same solution I thought of but wasn't sure!

> New changes LGTM, not merging myself though as it's still WIP?

WIP tag removed and merged.

PR #2446, opened 2020-06-02
Add date filter to documentation

### LINKED ISSUE #1931: Document date_filter parameter for timetable endpoint
**Is your feature request related to a problem? Please describe.**

The /timetable/personal endpoint accepts an undocumented `date_filter` query parameter in ISO 8601 format, i.e. YYYY-MM-DD

**Describe the solution you'd like**

Document this in uclapi.com/docs

zipy124:
## What does this PR do?
Adds the date filter to documentation, as you can get timetable for a specific day.

## ✅ Pull Request checklist

- [ ] Is this code complete?

no

## :squirrel: Deploy notes
Re-build front-end

## Anything else
Closes #1931


zipy124:
> In the code the name is actually `date_filter` but I agree that the name should be `date` (although not `Date`) as it'll be consistent with the bookings endpoint (the other endpoint we have that can filter by date). So this will require changes in the `views.py` and `app_helpers.py` as well.
> 
> I also think, in terms of consistency, optional parameters should be after the required ones, they seem to be placed in the middle AFAICT.
> 
> Otherwise, LGTM.

I've changed the parameter to 'date' however I haven't modified the variable names to that too, as I think date_filter is actually more verbose and fitting here. I've also re-ordered it in the documentation.

frontend/src/components/documentation/Routes/Timetable/GetPersonalTimetable.jsx
```diff
@@ -102,6 +102,12 @@ export default class GetPersonalTimetable extends React.Component {
                 requirement="required"
                 example="uclapi-5d58c3c4e6bf9c-c2910ad3b6e054-7ef60f44f1c14f-a05147bfd17fdb"
                 description="Authentication token." />
+              <Cell
+                name="Date"
```
Feverfew, line 106:
```suggestion
                name="date"
```
I'd also note this optional field is still before a required field.

backend/uclapi/timetable/views.py
```diff
@@ -66,7 +66,7 @@ def get_modules_timetable_endpoint(request, *args, **kwargs):
 
     modules = module_ids.split(',')
 
-    date_filter = request.GET.get("date_filter")
+    date_filter = request.GET.get("date")
```
Feverfew, line 69: The same must be done in `get_personal_timetable_endpoint()`. I'd also note that in that function they check for `KeyError`, which is unnecessary (I believe) if you do `s/request.GET[]/request.GET.get()`

Feverfew:
Also, please squash and merge this, should really only be one commit, instead of 12

zipy124:
> Also, please squash and merge this, should really only be one commit, instead of 12

I finally made squashing and merging the default after you raised this issue last time!

PR #2570, opened 2020-07-12
[TEST] PR to Test Hound

hueyy:
DO NOT MERGE

See https://github.com/uclapi/uclapi/pull/2570#issuecomment-657206130

### DIFF
--- a/frontend/.eslintrc
+++ b/frontend/.eslintrc
@@ -99,7 +99,7 @@
   },
   "settings": {
     "import/resolver": {
-      "webpack": "webpack.prod.js",
+      "webpack": "./nonsense.js",
       "alias": {
         "map": [
           [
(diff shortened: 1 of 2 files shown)


hueyy:
The problem is documented in this issue: https://github.com/houndci/hound/issues/1419

* We use a webpack import resolver to map import paths such as `./src/components/layout` to `Layout`
* Hound does not install dependencies, so it cannot resolve the paths to such imports

This problem can be resolved in the following ways:

* When we complete our frontend/backend separation, we might not need webpack, and we might be able to use an alternative import resolver such as [`eslint-import-resolver-alias`](https://www.npmjs.com/package/eslint-import-resolver-alias)
* We can simply live with long paths such as `./src/components/layout`.
* We can disable import resolution checking by setting `"import/extensions": "off" and "import/no-unresolved": "off"` in `.eslintrc`. If we want to continue checking import resolution locally, we can have a Hound-specific config with those checks disabled.

PR #2789, opened 2020-09-17
✂  Decouple warning & error pages from django

hueyy:
## What does this PR do?

The following 3 pages no longer require any `initial_data` supplied by Django. They're now static, as they should be.

* Warning - warns the user they are not completely logged out on all UCL services
* Error 404
* Error 500

## ✅ Pull Request checklist

- [x] Is this code complete?

No

## :squirrel: Deploy notes

Redeploy frontend

## Anything else

I'm not sure why codecov is failing even though the change in coverage is 0.00%.

### DIFF
--- a/backend/uclapi/dashboard/templates/404.html
+++ b/backend/uclapi/dashboard/templates/404.html
@@ -4,11 +4,11 @@
 
 <head>
 
-  <title>Warning | UCL API</title>
+  <title>Not Found | UCL API</title>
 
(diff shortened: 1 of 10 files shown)


zipy124 (approved):
Looks great, Will merge this and get it on staging ASAP. The reason codecov is complaining is because the lines added are not covered by tests, even though the lines they replace weren't either, hence the 0% change in coverage.

PR #2790, opened 2020-09-17
✂ Decouple dashboard from Django

hueyy:
## What does this PR do?

This allows the dashboard to exist as a static site on its own. Login and shibboleth is not yet done, but assuming a user is already logged in, a static version of the dashboard would 100% work.

Hound got upset whether I put the line break before or after the `+`, so I've ignored `W503` altogether.

I wrote a test for the new API I put in (`get_apps`), but test coverage for this part of the code is just low, hence the codecov fail.

## ✅ Pull Request checklist

- [x] Is this code complete?

No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,7 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
+/lib/
 lib64/
 parts/
 sdist/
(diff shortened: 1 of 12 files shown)


zipy124 (approved):
Everything looks good here, Will push to staging for further testing. W503 is because PEP8 changed to switch from + before the line break to + after the linebreak. Thus technically both have been both the pattern and anti-pattern at some point. The current new convention is to use it after the linebreak, as stated by donald knuth.

PR #2794, opened 2020-09-17
Handle timetable events with no bookings

Feverfew:
This patch handles timetable events that have no corresponding bookings
attached to them. This can occur during a "normal" year, but this is
especially prevalent during COVID-19 with online lectures dominating
people's timetables. Interestingly, this was already handled in
_get_timetable_events(), which is still used in some of the Python
only endpoints. Hence, the logic to handle this is adapted into parts
of the SQL and in the serialisation that still happens on the Python
side, where appropriate.

## What does this PR do?
For example: Changes the way we retrieve bookings.

## ✅ Pull Request checklist

- [ ] Is this code complete?

Yes/No

## :squirrel: Deploy notes

### DIFF
--- a/backend/uclapi/timetable/app_helpers.py
+++ b/backend/uclapi/timetable/app_helpers.py
@@ -1,6 +1,7 @@
 import datetime
 import json
 from distutils.util import strtobool
+from typing import List
 
 import redis
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


zipy124 (approved):
Everything looks good here, Will get this pushed to staging ASAP for testing along with everything else.

PR #2906, opened 2020-10-24
Use Cross-Env and update about page

greenfrogs:
## What does this PR do?
Add cross-env to frontend to allow webpack building with `npm start` on Windows. Created docker-compose and .env.example for database docker configuration. Updated about page.

## ✅ Pull Request checklist
- ✅ Is this code complete?
- ✳️ Are tests done/modified?
- ✳️ Are docs written for this PR?

No


hueyy:
Frontend changes look great, but what's the dev docker-compose file for?

greenfrogs:
> Frontend changes look great, but what's the dev docker-compose file for?

Docker-compose dev runs only the databases on docker and exposes the ports for connection from your local machine. This allows you to run the backend locally without needing to configure the databases. The .env file is configured similarly to `.env.docker.example`.

hueyy:
Ah, it so it's identical to the main docker-compose file but without the main app container. I suppose it's for those of us who want to run python & node outside of docker.

hueyy (approved):

backend/uclapi/.env.example
```diff
@@ -53,6 +54,7 @@ DB_CACHE_USERNAME=
 DB_CACHE_PASSWORD=
 DB_CACHE_HOST=
 DB_CACHE_PORT=5432
+DB_CACHE_POOL_SIZE=20
```
ILW8, line 57: This does not seem to be used anywhere, but I assume it's meant for [the update gencache script](https://github.com/uclapi/uclapi/blob/2da88489a11e6d1896a5baee69205da951c49cf7/backend/uclapi/timetable/management/commands/update_gencache.py#L333)? 
Same for DB_UCLAPI_POOL_SIZE and its respective file?

zipy124: This accurately represents the pools size we set at line 149 in settings.py, with 15 pool size and 5 for overflow, we planned I think on using the environment variable to set it, or we used to before we changed pooling mechanism maybe? Can safely be removed now at least I believe.

It does not reference the pool size in gencache.

PR #2941, opened 2020-11-08
Add ucl_groups to docs of userdata endpoint 

### LINKED ISSUE #2815: [BUG] ucl_groups reutnr list not documented in /docs/#oauth/user/data
Response part of the docs should note we give back a list, listing the LDAP groups that the user is a part of.

Feverfew:
## Summary
The docs don't mention this key despite it being in the response.
This commit adds documentation highlighting this fact.
Resolves #2815 
## Tests
None, change is pretty self-evident.


ChristopherHammond13 (commented):
I would avoid putting a multi-line comment inside a dictionary, otherwise makes sense.

Just first verify that nobody is using this data anyway. There are some larger users of UCL API that might just be using this field without us saying it's ok to do so.

zipy124:
as @ChristopherHammond13 said, I'm unsure if users might be using this so I'm hesitant to remove it. I'm also unsure how to check. Perhaps we can email all people who use the dashboard as a list to mail?

greenfrogs (approved):
LGTM

PR #2954, opened 2020-11-12
Workspaces tests and fixed bug in max_timestamp

greenfrogs:
## What does this PR do?
Adds tests for all workspaces api endpoints and fixes oversight in `cache_survey_sensors_max_timestamp` in `occupeye cache` causing it not to run resulting in no max_timestamp Redis entries to be created (and the sensors/lastupdated endpoint always returning invalid id).

## ✅ Pull Request checklist

- [x] Is this code complete?

Yes, only technically as before sensors/lastupdated always failed.

## :squirrel: Deploy notes
None


greenfrogs:
Superseded by #2968

PR #2966, opened 2020-11-14
Allow splitting analytics by service

### LINKED ISSUE #2814: [Feature Request] Separate most popular methods by service
In the `/dashboard/api/analytics/methods` endpoint we list the most popular methods. This however is not useful without also knowing the service for example both `https://uclapi.com/timetable/data/modules` and `https://uclapi.com/timetable/data/courses/modules` are classed as the `modules` method` but they are different.

ILW8:
## What does this PR do?
Allow splitting analytics by service so that different methods appear as separate entries in the analytics instead of being bunched up together as a single thing.

## ✅ Pull Request checklist

- [x] Is this code complete?

No; API clients need to explicitly specify a new request parameter for this to break existing API consumers.

## :squirrel: Deploy notes
None

## Anything else
Fixes #2814


ILW8:
Not too sure about the documentation.

zipy124 (approved):
Everything looks great here.

PR #2967, opened 2020-11-14
Increase Max Line Length to 119

greenfrogs:
## What does this PR do?
Increases the max line length from 79 to 119 for flake8. It also includes examples for the workspaces/commands folder.

Historically the line length of 79 was chosen in PEP8 for comparability with systems and terminals which could only display 80x24 characters. This is still common in C style checkers, especially for some legacy systems where this may cause a problem (for example with punch-cards). However unlike C, Python is OOP which commonly results in lots of classes and to make them understandable longer descriptive names are preferred.

With the sort of modern systems UCL API are designing there is no need for this backwards support. Windows terminal now defaults to 120 characters and even PyCharm which follows almost all other PEP8 standards uses 120 characters for Python. One of the few options in the infamously uncompromising black Python code formatter is line length (and the default is longer than PEP8). Developers can easily understand 120 characters at once and the examples below highlight just how unreadable some of the code base (especially tests) has become.

In summary, UCL API is trying to write code which is readable and understandable, not that can be written on a punch card. 

### Line Length 79
before

### Line Length 119
after


ILW8 (commented):
Agree, I've had some run-in with style checker earlier about lines I thought was of perfectly reasonable length

ChristopherHammond13:
Did not know you could do this -- awesome! Just be aware though that editing side by side, especially on smaller screens, will become impossible very quickly. If that's something you're willing to forgo then this is smart. Remember, however, that you can always do this to temporarily bypass the limit:
```
very very very very long line  # noqa
```

Feverfew (approved):
I agree with the idea. I usually develop on a laptop and prefer the 80 tbh, but I can always enable line-wrap in my editor/IDE, the inverse can't be done as easily ("uncompress" from 80 -> 120).

Please separate into two separate commits, i.e. one for both files.

Feverfew (approved):
LGTM will squash and merge

PR #3033, opened 2020-12-12
Disable webhooks

ILW8:
## What does this PR do?
Disables webhook

## ✅ Pull Request checklist

- [ x ] Is this code complete?
- [ N/A ] Are tests done/modified?

No, but webhooks are gone.

## Anything else
Make sure that nobody actually uses webhook/warn them of this being disabled. 
This PR currently does not clean up the webhook code, it simply comments out the line that calls executing the webhook code in `update_gencache`. Awaiting feedback.

### DIFF
--- a/backend/uclapi/timetable/management/commands/update_gencache.py
+++ b/backend/uclapi/timetable/management/commands/update_gencache.py
@@ -383,7 +383,7 @@ class Command(BaseCommand):
             else:
                 print("Could not find appropriate incident in Cachet!")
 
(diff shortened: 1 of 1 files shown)


zipy124:
As webhooks send back the status code, an easy way for us to check usage is simply to see how many 200's we get back.

Feverfew:
> As webhooks send back the status code, an easy way for us to check usage is simply to see how many 200's we get back.

Let's do that then and decide from there.

PR #3114, opened 2021-01-12
✨ NEW: use GitHub Action for CI/CD

### LINKED ISSUE #2955: [MIGRATION] travis-ci.org to Github Actions before Brexit
**Describe the  migration**
Travis-ci.org is being shutdown on December 31st, 2020. Repositories need to be migrated to travis-ci.com:
https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom

https://docs.travis-ci.com/user/migrate/open-source-repository-migration

hueyy:
## What does this PR do?
Resolves #2955.

This does CI only. Deployment to staging is untested because I didn't want to pollute staging or master. I'll pollute staging and test CD in a different PR later.

## ✅ Pull Request checklist

- [x] Is this code complete?

No

## :squirrel: Deploy notes

## Anything else

Please **squash** and merge. I put a lot of commits into this PR for testing purposes.


Feverfew (dismissed):
LGTM.

I guess remove `travis.yml` as well (and any other associated CI files).

In addition,, should probably delete everything that we have on Travis for good measure.

Wait on Zak to check this as well.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,95 @@
+name: CI
+
+on:
+  push
```
greenfrogs, line 4: Do we want to run CI on pull requests too?
on: [push, pull_request]

.github/workflows/CD.yml
```diff
@@ -0,0 +1,83 @@
+name: CD
+
+on:
+  workflow_dispatch:
+    inputs:
+      confirmation:
+        description: 'Are you sure you want to deploy master to staging? (type yes below if so)'     
+        required: true
+        default: 'yes'
+  push:
```
greenfrogs, line 10: GitHub actions can run concurrently, and this runs on any push to master. Is it worthwhile using something like [turnstyle](https://github.com/softprops/turnstyle) (not sure if this is natively supported yet) to prevent running two different deployments at once?

.github/workflows/CI.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: build frontend
+        run: |
+          cd ./frontend && \
+          npm run build
+
+      - name: run our tests and check test coverage
+        run: |
+          cd ./backend/uclapi && \
+          coverage run --source='.' --omit='*migrations*' manage.py test --testrunner 'uclapi.custom_test_runner.NoDbTestRunner' --settings=uclapi.settings_mocked && \
+          codecov
```
greenfrogs, line 95: It might be better to transition to using the Github Action:
[https://github.com/codecov/codecov-action](https://github.com/codecov/codecov-action)

greenfrogs (changes requested):
Looks good but see comments

.github/workflows/CI.yml
hueyy: All PRs have pushes, so afaik, this should run on PRs as well. No need to run it twice. Unless there's some edge case I'm missing?

.github/workflows/CI.yml
hueyy: Okay.

.github/workflows/CD.yml
hueyy: This makes sense. Doesn't seem to be natively supported yet so I'll put turnstyle in.

.github/workflows/CI.yml
hueyy: Good points. Will make the changes.

.github/workflows/CD.yml
```diff
@@ -0,0 +1,91 @@
+name: CD
+
+on:
+  pull_request:
+    types: [closed]
```
greenfrogs, line 5: I am not sure closed is appropriate as it may not have been merged, I have my personal deployments set on push, master, either with an if inside a job or on for the entire file:
`
on:
  push:
    branches:
      - master

if: github.event_name == 'push'
needs: push
`

greenfrogs: Sorry I made this confusing, I think on needs the push with branch master. At the moment it will only run on a pull_request event type and then we check if github.event_name (which can only ever be pull request) is equal to push.

hueyy: No, it's my fault. I wanted to avoid the CD workflow running twice when a PR is merged into master (once for the PR event, and another for the push event), but I had a brainfart and messed up the job conditional. I think this should do it.

It won't run for direct pushes to master, but I don't think we should be doing that anyway. All changes should come in the form of merged PRs. If Zak needs to urgently push a change to master, he can deploy the frontend manually anyway.

greenfrogs (approved):

Issue #3060, opened 2020-12-20
[BUG] Add Types to our code


Feverfew:
**Is your feature request related to a problem? Please describe.**
Dynamic languages are cool and all, but make it hard to figure out what possible types there could be statically. This causes countless bugs, especially as our gencache DB is nullable everywhere!

**Describe the solution you'd like**
Type check all the code, with CI to prevent type-less code!

**Describe alternatives you've considered (optional)**
Do nothing

**Additional context(optional)**
This is quite a daunting project, and there are doubts this will be feasible with Django but we need to experiment before we write it off. We should aim to type check the `common` and `uclapi` folders first, as a lot of other apps depend on them. After that we can choose whatever is easiest, just so that we get used to it. In addiiton we can use MonkeyType to do a lot of the work for us! We should probably get CI involved quite soon, so it will check all folders that we know are typed, to make sure we don't regress!

https://mypy.readthedocs.io/en/stable/
https://github.com/Instagram/MonkeyType
https://github.com/typeddjango
https://sobolevn.me/2019/08/typechecking-django-and-drf
https://instagram-engineering.com/let-your-code-type-hint-itself-introducing-open-source-monkeytype-a855c7284881
https://instagram-engineering.com/static-analysis-at-scale-an-instagram-story-8f498ab71a0c

PR #3118, opened 2021-01-14
Replace docs with auto-generated OpenAPI docs

### LINKED ISSUE #2964: [BUG] Inconsistency with trailing slashes with API URLs in documentation sample code, 404 when trailing slash is present
**Describe the bug**
The URLs provided in the sample code in the API Documentation [sometimes contain trailing slashes](https://uclapi.com/docs/#dashboard/api/analytics/services). When running the sample code, we would get 404 because of the trailing slash, whereas it would work if the trailing slash is removed from the URL.

**To Reproduce**
Steps to reproduce the behavior (using one of the endpoints):
1. request https://uclapi.com/dashboard/api/analytics/methods/
2. Get 404

**Expected behavior**
A response from the API (not a 404)

**Additional context**
Django usually uses trailing slashes with all their URLs, however there is significant controversy regarding whether an URI with a trailing slash is the same as another URI _without_ a trailing slash ([discussion 1](https://stackoverflow.com/questions/61547014/restful-uri-trailing-slash-or-no-trailing-slash), [discussion 2](https://github.com/open-policy-agent/opa/issues/972#issuecomment-427937639)). 

By default, Django [has a setting that redirects URLs](https://docs.djangoproject.com/en/dev/ref/settings/?from=olddocs#append-slash) that don't match any urlconf to the same URL plus a trailing slash. Adding a trailing slash to the URL patterns in the respective `urls.py` would likely solve the issue, however the documentation states that:
> Note that the redirect **may** cause any data submitted in a POST request to be lost.

and I'm not sure how that affects the endpoints that do use POST. @greenfrogs

### LINKED ISSUE #3054: OpenAPI Schema
**Is your feature request related to a problem? Please describe.**
Currently we have to edit the frontend manually to create docs, which offers no checking for us and is quite tedious.

**Describe the solution you'd like**
OpenAPI schema in this repo!

**Describe alternatives you've considered (optional)**
The status quo.

### LINKED ISSUE #3055: [Feature Request] Auto-generate documentation with OpenAPI
**Is your feature request related to a problem? Please describe.**
Writing documentation for APIs is annoying...

**Describe the solution you'd like**
Statically generate the documentation based on the schema (yay no javascript ;) ).

**Describe alternatives you've considered (optional)**
The status quo, it sucks.

shu8:
## What does this PR do?
Replaces the manually-written documentation with automatically Swagger-rendered docs from our OpenAPI specification at https://github.com/uclapi/uclapi-openapi/blob/master/uclapi.json.

Unless I've missed anything, the new Swagger docs should have everything that the old docs had **except for** the example snippets for each endpoint.

The old example snippets used to be mostly the same for each endpoint, so if we wanted to keep this, maybe we could just add a single example for each type at the top?

A cURL example is automatically generated by Swagger when you try the API via the UI though.

image

image

Closes #3055. Closes #3054.

## ✅ Pull Request checklist

- [ ] Is this code complete?

No

## :squirrel: Deploy notes
N/A

## Anything else
Fixes #2964


hueyy:
Some of the code examples were a bit special, like the OAuth stuff, which I think it might be useful to keep. Is there any way we can include some of those code snippets, maybe in the description or something?

Feverfew:
> Some of the code examples were a bit special, like the OAuth stuff, which I think it might be useful to keep. Is there any way we can include some of those code snippets, maybe in the description of each endpoint or something?

I agree the OAuth workflow should be something we can keep.

A couple of questions I'd like clarified:
1) One thing is that the SDKs should always correspond to what is currently running on the server(s), so the deploy scripts for staging and production should be where the new SDK is pushed from. This sounds like it requires two SDKs, as the code on staging and production is different. Is it possible to take this into account?
2) Related to the above, I'd prefer if the definition was in this repo (and sounds like it will make it easier to take into account the first point). Firstly, this will mean that its easier to enforce that changes in the API are also reflected in changes to the OpenAPI definition at the same time (i.e. in the same PR). Secondly, it sounds like that it will make it easier to meet the requirements of the point I made above.

Otherwise, I like the fact that we've got rid of this many lines of code!

Could we also also have documentation in the `backend/source` directory to explain the recommended workflow in regards to editing the schema?

shu8:
I've updated the spec to add code samples for the OAuth endpoints, and added one example at the top for how to call the roombookings api, which hopefully people will find easy to tweak for the other endpoints: https://github.com/uclapi/uclapi-openapi/commit/c79149fa4148d5c47e61d764f1b59c5db6941cc8

---

I think adding the spec file to this repo makes sense for both the reasons you mentioned, but just a quick (possibly stupid) question: how/when do we currently promote staging to production?

Having separate SDKs should also work, and it might be enough to just create separate branches in the SDK repos and publishing to npm, etc. with different version numbers depending on which one it is.

Feverfew:
> I think adding the spec file to this repo makes sense for both the reasons you mentioned, but just a quick (possibly stupid) question: how/when do we currently promote staging to production?

Whenever Zak feels like it, manually. Should be automated with Github Actions, after integration tests on staging are run.

> Having separate SDKs should also work, and it might be enough to just create separate branches in the SDK repos and publishing to npm, etc. with different version numbers depending on which one it is.
Yeah the workflow I'm thinking of is.
1. Master is updated.
2. Deployment to staging. Once deployed, job is triggered to create new staging SDK(s) based on OpenAPI spec in the repo and push them to NPM automatically (potentially with some smoke tests?).
3. Same for production, but obviously a different SDK. Whether it's a different branch or repo, I'm not too worried about.

Feverfew:
As discussed in the meeting once the OpenAPI spec is added to this repo and the documentation is built from it, then I'm happy to review and merge this. We can deal with SDK stuff in a separate PR

frontend/src/pages/Documentation.jsx
```diff
@@ -1,58 +1,17 @@
...
 
 const Documentation = () => (
-  <StylesProvider injectFirst>
-    <MuiThemeProvider theme={muiTheme}>
-      <DocumentationComponent />
-    </MuiThemeProvider>
-  </StylesProvider>
+  <div className='vertical-padding'>
+    <NavBar isScroll={false} />
+    <SwaggerUI url="https://cdn.jsdelivr.net/gh/uclapi/uclapi-openapi/uclapi.json" />
```
Feverfew, line 13: Can this be built statically? The reason I ask is, firstly for performance reasons, but more importantly is for consistency. At what point is `https://cdn.jsdelivr.net/gh/uclapi/uclapi-openapi/uclapi.json` updated? It seems like either the documentation should be generated as part of a build step or be part of the statics?

Feverfew (changes requested):
One question that I'd like to hash out before approval

frontend/src/pages/Documentation.jsx
shu8: Yeah I agree statically building is best -- I should have really marked this as WIP after our meeting sorry! 

I haven't had much time to look into the best way to build it statically and join it with the webpack build yet, but I'll try looking into it ASAP - openapi-generator has a [dynamic-html docs generation option](https://openapi-generator.tech/docs/generators/dynamic-html) which looks promising for us!

shu8: It's part of the build process now!

shu8:
@Feverfew I've:

- added some sphinx docs for the openapi documentation process,
- moved the openapi docs rendering to be a build-time step so it is always in sync with the current deployment, and
- updated the spec to add descriptions to all the fields (which we had in the existing docs but I forgot to copy over when I did the openapi spec originally!)

if it looks good then I will move onto the SDK deployment work and archive the old `uclapi-openapi` repo (https://github.com/uclapi/uclapi-openapi)!

Feverfew (dismissed):
LGTM let's get this in staging.

greenfrogs:
We need to wait for staging to be fixed first, and #3052 to merge. Possibly also adding the new docs from #3052 (otherwise we are going to have some really annoying merge conflicts on the other request).

shu8:
> Possibly also adding the new docs from #3052 (otherwise we are going to have some really annoying merge conflicts on the other request).

I'll add those into this PR now, so once workspaces is merged we can get this in quicker!

shu8:
I've added the docs from #3052 into this PR but the review got dismissed with the new push! Please could either of you have another quick look @greenfrogs @Feverfew!

shu8:
sorry for the delay but all merge conflicts and tests should be fixed now for the final re-review! @greenfrogs @Feverfew

zipy124:
@uclapi/ucl-api-team are we ready to get this into master now?

shu8:
@zipy124 I think it is -- the last review got dismissed because of all the merge conflicts I fixed and the new workspaces docs but nothing else has changed! 

I'll resolve this package-lock.json conflict now too!

uclapi.openapi.json
```diff
@@ -0,0 +1,4256 @@
...
+    "contact" : {
+      "name" : "API Support",
+      "email" : "isd.apiteam@ucl.ac.uk",
+      "url" : "https://uclapi.com"
+    },
+    "version" : "1.0.0-oas3",
```
greenfrogs, line 11: nit: do we want to match this version to some sort of release of the API?

uclapi.openapi.json
```diff
@@ -0,0 +1,4256 @@
...
+            "description": "Human readable department name that owns the module."
+          },
+          "department_id" : {
+            "type" : "string",
+            "example" : "COMPS_ENG",
+            "description": "Department code that this module comes under."
+          },
+          "email" : {
+            "type" : "string",
+            "example" : "ucachun@ucl.ac.yk",
```
greenfrogs, line 576: nit: ac.uk

uclapi.openapi.json
```diff
@@ -0,0 +1,4256 @@
...
+            "description": "Name of the site / building where the event is happening."
+          },
+          "type" : {
+            "type" : "string",
+            "example" : "CB",
+            "description": "Type of the room eg., Centrally Bookable (CB) or Departmentally Bookable (DB)."
+          },
+          "capacity" : {
+            "type" : "integer",
+            "format" : "int64",
```
greenfrogs, line 610: ooc: any reason for `int64` instead of `number`?

uclapi.openapi.json
```diff
@@ -0,0 +1,4256 @@
...
+                "type" : "object",
+                "properties" : {
+                  "lat" : {
+                    "type" : "number",
+                    "example" : "51.524699",
+                    "description": "Latitude of location"
+                  },
+                  "lng" : {
+                    "type" : "number",
+                    "example" : "-0.13366",
```
greenfrogs, line 729: nit: can these numbers be outside of quotes like above (or do floats cause problems)?

uclapi.openapi.json
```diff
@@ -0,0 +1,4256 @@
...
+            "description": "Describes what the booking is. Could contain a module code (for example WIBRG005) or just the type of activity (for example Lecture)."
+          },
+          "roomname" : {
+            "type" : "string",
+            "example" : "Torrington (1-19) 433",
+            "description": "The name of the room. It often includes the name of the site (building) as well."
+          },
+          "siteid" : {
+            "type" : "number",
+            "example" : 86.0,
```
greenfrogs, line 762: super nit: remove .0

uclapi.openapi.json
```diff
@@ -0,0 +1,4256 @@
...
+            "description": "The name of the person who made the booking. Substrings of the contact name can be used: Queries for Mark will include Mark Herbster. Sometimes, a society or student group may be the contact for a booking."
+          },
+          "weeknumber" : {
+            "type" : "number",
+            "example" : 1,
+            "description": "Timetabled week number."
+          },
+          "roomid" : {
+            "type" : "number",
+            "example" : "433",
```
greenfrogs, line 777: nit: number in quotes

uclapi.openapi.json
```diff
@@ -0,0 +1,4256 @@
...
+                "description": "Object containing a latitude and longitude.",
+                "properties" : {
+                  "lat" : {
+                    "type" : "string",
+                    "example" : "51.522897",
+                    "description": "Latitude of the survey location."
+                  },
+                  "lng" : {
+                    "type" : "string",
+                    "example" : "-0.127864",
```
greenfrogs, line 947: nit: number in quotes

uclapi.openapi.json
```diff
@@ -0,0 +1,4256 @@
...
+            "description": "A unique ID for the sensor."
+          },
+          "pir_address" : {
+            "type" : "string",
+            "example" : "1",
+            "description": "PIR address."
+          },
+          "last_trigger_timestamp" : {
+            "type" : "string",
+            "example" : "2018-02-15T22:42:28+00:00",
```
greenfrogs, line 1111: ooc: do we return time zone here? Our API seems to be really inconsistent 😅

uclapi.openapi.json
```diff
@@ -0,0 +1,4256 @@
...
+            "description": "The ID of this survey device."
+          },
+          "share_id" : {
+            "type" : "string",
+            "example" : "None",
+            "description": "The Share ID."
+          },
+          "x_pos" : {
+            "type" : "string",
+            "example" : "32432.0",
```
greenfrogs, line 1092: This might be a mistake in the API but does this return a string rather than a number? (same for `survey_device_id`)

uclapi.openapi.json
```diff
@@ -0,0 +1,4256 @@
...
+            "additionalProperties" : {
+              "$ref" : "#/components/schemas/average"
+            }
+          }
+        }
+      },
+      "historical_survey_data" : {
+        "properties" : {
+          "sensor_id" : {
+            "type" : "integer",
```
greenfrogs, line 1144: nit: integer instead of number (same with `averages`)

uclapi.openapi.json
```diff
@@ -0,0 +1,4256 @@
...
+            "example": 1,
+            "description": "The survey device ID of the sensor."
+          }
+        }
+      },
+      "historical_survey" : {
+        "properties" : {
+          "survey_id" : {
+            "type" : "integer",
```
greenfrogs, line 1187: nit: int

uclapi.openapi.json
```diff
@@ -0,0 +1,4256 @@
...
+                "xml" : {
+                  "attribute" : true
+                }
+              },
+              "image" : {
+                "type" : "object",
+                "properties" : {
+                  "width" : {
+                    "type" : "string",
+                    "example" : "28329.0",
```
greenfrogs, line 1340: nit: same as above string for number value

uclapi.openapi.json
```diff
@@ -0,0 +1,4256 @@
...
+        "tags" : [ "OAuth" ],
+        "parameters" : [ {
+          "name" : "client_id",
+          "in" : "query",
+          "description" : "Client ID of the authenticating app.",
+          "required" : true,
+          "schema" : {
+            "type" : "string",
+            "example" : 1
```
greenfrogs, line 2137: nit: is this a string? is this a good example?

uclapi.openapi.json
```diff
@@ -0,0 +1,4256 @@
...
+                  "Invalid parameters" : {
+                    "value" : {
+                      "ok" : false,
+                      "error" : "start_datetime or end_datetime not provided"
+                    }
+                  },
+                  "Invalid date/time" : {
+                    "value" : {
+                      "ok" : false,
+                      "error" : "date/time isn't formatted as suggested in the docs"
```
greenfrogs, line 2795: nit: I think this is a mistake in the API, I believe we format date/time as datetime

uclapi.openapi.json
```diff
@@ -0,0 +1,4256 @@
...
+              }
+            }
+          }
+        }
+      }
+    }
+  },
+  "servers" : [ {
+    "url" : "https://uclapi.com",
+    "description" : "Deployment server"
```
greenfrogs, line 4251: nit: can we call this `Production server`?

greenfrogs (approved):
@shu8 sorry for the very late review. This is an excellent piece of work and will make it so much easier to support and use the API in the future.

I have some really small comments but they are mostly only nits so I am happy to approve for merging when you feel it is best. Lots of my comments are likely problems with the underlying API or weird inconsistencies which have risen as a result of difficulty in seeing datatypes/phrasing for the entire API. Feel free to not address these and I can raise issues to address them in the future.

shu8:
Thanks @greenfrogs! Yeah those all seem like things we should fix!

I should have time to go over these (and go through all the endpoints fully once more to see if there is any other weirdness!) over the next week but it might be good to get this into master/staging and I can submit another PR for all the fixes?

zipy124:
> Thanks @greenfrogs! Yeah those all seem like things we should fix!
> 
> I should have time to go over these (and go through all the endpoints fully once more to see if there is any other weirdness!) over the next week but it might be good to get this into master/staging and I can submit another PR for all the fixes?

I'm good with this decision!

uclapi.openapi.json
shu8: Yeah it look like we return exactly what comes from occupeye

uclapi.openapi.json
shu8: Turns out we actually return lat/long as a string! that seems intentional, maybe for storage or precisions concerns?

uclapi.openapi.json
shu8: this is also actually a string, updated!

uclapi.openapi.json
shu8: Yes! We can do that along with #3379!

uclapi.openapi.json
shu8: Yeah looks like we're returning a string in the API, again using the value straight from Occupeye

uclapi.openapi.json
shu8: I think it might be better to be more specific and stick with `integer` here?

uclapi.openapi.json
shu8: same with this! what do you think?

uclapi.openapi.json
shu8: I think this is a string because it's a HTML SVG so all the attributes would be strings

uclapi.openapi.json
shu8: updated error message to be more clear: `start_datetime/end_datetime isn't formatted as suggested in the docs`

uclapi.openapi.json
shu8: Looking through the code we have a lot of fields with types that I don't really get, e.g., `weeknumber` is a `FloatField` for `/roombookings`, and `BigIntegerField` for `/timetable`.

we could probably update a lot of our types -- what do you think? if we did #3060 this could go along with it.

PR #3133, opened 2021-01-30
🐛 Status code from error_404 and error_500 views

greenfrogs:
## What does this PR do?
Fix error_404 and error_500 to return correct error codes.

No


Feverfew (approved):
LGTM

Issue #3053, opened 2020-12-20
[Feature Request] Utilize Sentry more effectively


Feverfew:
We need to do the following:

1) Upgrade our plan to use Sentry for Education
2) Migrate from raven client (deprecated) to sentry client, enabling it also for our management commands that are tied to a crontab.
3) Enable Django (and Github?) integration.

zipy124:
2 and 3 are doable. 1 we technically do not qualify as we aren't acting as students, teachers or researchers and so we will pay for it instead and get a buisness account maybe?

greenfrogs:
Add:
- [ ] Gencache errors
- [ ] Add release information to Sentry (wait for #3114)

PR #3134, opened 2021-01-31
✨ Migrate to new Sentry API

greenfrogs:
## What does this PR do?
Migrates from Raven to the new Sentry API for Django and React. Part of #3053

## ✅ Pull Request checklist
- [x] Initial migration
- [x] Scrub personal data
- [ ] Gencache errors
- [ ] Add release information to Sentry (wait for #3114)

No

## :squirrel: Deploy notes
Update new DSN keys for Django in .env and add new SENTRY_DSN_REACT in Travis for webpack.


backend/uclapi/requirements.txt
```diff
@@ -27,7 +27,7 @@ psycopg2-binary==2.8.6
 python-dateutil==2.8.1
 pymsteams==0.1.14
 pytz==2020.4
-raven==6.10.0
+sentry-sdk==0.19.5
```
Feverfew, line 30:
```suggestion
sentry-sdk==0.20.1
```
Benefits of not reviewing... although test that it doesn't break anything :P

backend/uclapi/uclapi/ourcelery.py
```diff
@@ -17,9 +14,15 @@
...
-            raven_register_logger_signal(client)
-            raven_register_signal(client)
+            import sentry_sdk
+            from sentry_sdk.integrations.django import DjangoIntegration
+
+            sentry_sdk.init(
+                dsn=os.environ.get("SENTRY_DSN"),
+                integrations=[DjangoIntegration()],
+                traces_sample_rate=0.01,
+                send_default_pii=True
```
Feverfew, line 24: We need the `before_send` keyword argument to be filled as well: https://docs.sentry.io/platforms/python/configuration/filtering/#using-beforesend

This is particularly important for us as we have the legacy of putting confidential information as a query parameter, which we should prevent from being stored on other servers. So make sure we strip tokens, client_id/state etc.

Also strip anything else that we don't need for debugging but is sensitive.

backend/uclapi/uclapi/settings.py
```diff
@@ -154,9 +153,16 @@
...
-}
+if os.environ.get('SENTRY_DSN'):
+    import sentry_sdk
+    from sentry_sdk.integrations.django import DjangoIntegration
+
+    sentry_sdk.init(
+        dsn=os.environ.get('SENTRY_DSN'),
+        integrations=[DjangoIntegration()],
+        traces_sample_rate=0.01,
+        send_default_pii=True
```
Feverfew, line 164: does this work for us? We don't use the default Django authentication stuff.

frontend/src/pages/Warning.jsx
```diff
@@ -1,3 +1,5 @@
+import '../lib/ErrorReporting'
```
Feverfew, line 1: Anyway that this boilerplate can be avoided.

Feverfew (changes requested):
Fairly straightforward. Just a few questions I'd like answered.

Await review from Huey in terms of frontend and Zak so that he puts in the required stuff in the `.env`.

backend/uclapi/uclapi/settings.py
greenfrogs: We don't get the Django auth benefit however this sends us additional information about the user client (such as ip address and user agent). In the interest of privacy I have disabled it.

frontend/src/pages/Warning.jsx
greenfrogs: This was the shortest I could make it, unless @hueyy has any ideas. From my understand each page is an entry point so we need it on each of them.

backend/uclapi/uclapi/ourcelery.py
greenfrogs: It turns out that sentry sends quite a bit more information than I was expecting, some in several different forms (I counted the URL with get parameters provided in at least 5 different nodes of the tree). Sentry has some server side scrubbing but ideally we would not send anything sensitive.

This is made harder by the event tree being very variable and has resulted in this recursive explorer checking for dict/list entries. We then check against a list of known keys which contain sensitive information and remove those entries. For the URL entries and some trace backs (the get parameters are passed in as arguments so sometimes shown in the stack trace), this approach uses regex's to match the possibly generated tokens, client ids, and client secrets. I am not sure of a better solution (without regex) which won't redact too much debugging information.

frontend/src/pages/Warning.jsx
hueyy: Yeah the way it's set up right now requires this. Not worth doing something fancy like injecting it into the React dependency via webpack imo.

hueyy (approved):
LGTM!

backend/uclapi/uclapi/settings.py
```diff
@@ -154,9 +155,38 @@
...
-RAVEN_CONFIG = {
-    'dsn': os.environ.get("SENTRY_DSN"),
-}
+if os.environ.get('SENTRY_DSN'):
+    import sentry_sdk
+    from sentry_sdk.integrations.django import DjangoIntegration
+    from sentry_sdk.integrations.redis import RedisIntegration
+
+    def remove_token(event, _):
+        scrubbers_keys = ['token', 'client_secret', 'X-RateLimit-Remaining', 'X-RateLimit-Limit',
```
Feverfew, line 164: Lets add in `code` and `client_id` for good measure.

Feverfew (approved):
Otherwise LGTM, Let Zak merge so that he updates the `.env`.

greenfrogs:
Please do not merge until I have added the gencache errors and added deployment to GitHub Actions.

Feverfew:
> Please do not merge until I have added the gencache errors and added deployment to GitHub Actions.

Marked as draft :)

zipy124 (approved):

PR #3175, opened 2021-02-16
🚀 Use AWS_S3_STATICS_CREDENTIALS_ENABLED

greenfrogs:
## What does this PR do?
Fixes a very very long running bug where we never actually set any AWS secrets.


greenfrogs:
This might not be required as UCL API staging is a public bucket:
[https://github.com/jschneier/django-storages/issues/254](https://github.com/jschneier/django-storages/issues/254)

hueyy:
LMAO good find

ILW8 (approved):
How many times do you live? Let's give it a shot

PR #3176, opened 2021-02-16
🚀 AWS bucket is public do not need auth

greenfrogs:
## What does this PR do?
🚀 AWS bucket is public do not need authentication


ILW8 (approved):

greenfrogs:
Please see, [ttps://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html](ttps://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html) and AWS_QUERYSTRING_AUTH:

`
Setting AWS_QUERYSTRING_AUTH to False to remove query parameter authentication from generated URLs. This can be useful if your S3 buckets are public.
`

PR #3177, opened 2021-02-16
🚀 Provide env to frontend build

greenfrogs:
## What does this PR do?
🚀 Provide env to frontend build. This is also possibly the cause of the production using the staging bucket.


ILW8 (approved):
Fingers crossed

PR #3197, opened 2021-02-21
✨ Optionally Allow Bearer Authentication

greenfrogs:
## What does this PR do?
In the long term it will be better to allow bearer authentication along with the traditional query or multipart form currently supported. This is very useful for OpenAPI and the auto filling of tokens, #3118. Bearer auth also has the advantage that we are less likely (via Django or Sentry) log any secrets.

The current form of a bearer auth is:
uclapi-token-here,client-secret-here

A standard bearer token also includes a Bearer or Authentication prefix which is optional:
Bearer uclapi-token-here,client-secret-here

For endpoints without personal data only the uclapi-token needs to be included:
Bearer uclapi-token-here     OR     uclapi-token-here

For example:
Authorization: Bearer uclapi-e41308262f23egc-8069e4a41d8a4d7-0d3911b9342ec5b-65b6150c1bd0d49

In the far distance future this may lay the groundwork for the deprecation of get parameters.

## ✅ Pull Request checklist

- [x] Is this code complete?

No

### DIFF
--- a/backend/uclapi/common/decorators.py
+++ b/backend/uclapi/common/decorators.py
@@ -305,7 +305,18 @@ def uclapi_protected_endpoint(
                     raise UclApiIncorrectDecoratorUsageException
 
             # In any case, a token should be provided
-            token_code = get_var(request, 'token')
+            if 'HTTP_AUTHORIZATION' in request.META:
+                token_code = request.META['HTTP_AUTHORIZATION']
+
+                # To remove the starting BEARER or AUTHENTICATION if provided
+                if ' ' in token_code:
+                    token_code = token_code.split(' ')[1]
+
+                # If multiple tokens provided, for example client_secret
+                if ',' in token_code:
+                    token_code = token_code.split(',')[0]
+            else:
+                token_code = get_var(request, 'token')
             if token_code is None:
                 response = JsonResponse({
                     "ok": False,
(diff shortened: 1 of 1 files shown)


backend/uclapi/common/decorators.py
```diff
@@ -305,7 +305,18 @@ def wrapped(request, *args, **kwargs):
                     raise UclApiIncorrectDecoratorUsageException
 
             # In any case, a token should be provided
-            token_code = get_var(request, 'token')
+            if 'HTTP_AUTHORIZATION' in request.META:
+                token_code = request.META['HTTP_AUTHORIZATION']
+
+                # To remove the starting BEARER or AUTHENTICATION if provided
+                if ' ' in token_code:
+                    token_code = token_code.split(' ')[1]
```
Feverfew, line 313:
```suggestion
                token_code = token_code.lstrip('Bearer: ')
```
Firstly, this will match anything before. Secondly, I thought only `Bearer: <token>` was valid, with a quick google I didn't see any mention of `Authentication: <token>`.

backend/uclapi/common/decorators.py
```diff
@@ -316,7 +327,23 @@ def wrapped(request, *args, **kwargs):
...
             if token_code.startswith('uclapi-user-'):
                 # The token is an OAuth token, so apply OAuth logic
-                client_secret = get_var(request, 'client_secret')
+
+                if 'HTTP_AUTHORIZATION' in request.META:
+                    client_secret = request.META['HTTP_AUTHORIZATION']
+
+                    # To remove the starting BEARER or AUTHENTICATION if provided
+                    if ' ' in token_code:
+                        client_secret = client_secret.split(' ')[1]
```
Feverfew, line 336:
```suggestion
                        client_secret = client_secret.lstrip('Bearer: ')
```

Feverfew (changes requested):
Before I approve this the following needs to be done:

- [ ] Docs updated accordingly.
- [ ] Deprecation warnings added.
- [ ] Tests needed.

The first two can wait until #3118 is merged though.

backend/uclapi/common/decorators.py
greenfrogs: Yes you are absolutely right, just read through the [RFC-6750](https://tools.ietf.org/html/rfc6750) and not sure where I got the idea it could be any string.

ILW8: I believe [RFC6750 section 2.1](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1) specifies a single space character between `Bearer` and the actual token.

Suhas-13:
Can I get assigned this?

PR #3265, opened 2021-03-30
🐛 CI Use Settings When Running Migrate

greenfrogs:
## What does this PR do?
Follow up from #3253 trying to fix the same bug, migrate originally was not loading settings from uclapi.settings.

## ✅ Pull Request checklist

- [X] Is this code complete?

No


ILW8 (approved):
Let's hope this works

PR #3388, opened 2021-06-26
Use .env file variables in docker-compose

hueyy:
## What does this PR do?

Currently, our `docker-compose.yml` file hardcodes a number of environment variables. The variables specified in the `environment` key override those specified in the `.env` file, which can be a cause of unexpected behaviour. For instance, changing `UCLAPI_DOMAIN` or `UCLAPI_PRODUCTION` in the env file would previously have no effect when running the uclapi in docker-compose.

This PR hardcodes in the docker-compose file only the values that are already hardcoded elsewhere in that same file (namely the postgres/redis host, username, password, etc.), and leaves the rest of the environment variables to be specified by `backend/uclapi/.env`, with default values set by `backend/uclapi/.env.example`.

This PR also updates `backend/uclapi/.env.docker.example` to use values that work by default in docker-compose.

## ✅ Pull Request checklist

- [x] Is this code complete?

No

## 🚀 Deploy notes

This has no impact on deployment.

### DIFF
--- a/backend/uclapi/.env.docker.example
+++ b/backend/uclapi/.env.docker.example
@@ -2,20 +2,20 @@
 ## These settings are for the app as a whole, and not for any individual components.
 
 # Secret key for Django to use. Auto-generate this and keep it secret!
-SECRET_KEY=
+SECRET_KEY=iRlsWf2BF2nnNy1zIsZcun6fZX5GGpk0
 # Root of Shibboleth's authentication. The example below might apply to fakeshibboleth;
 # an actual example would be https://uclapi.com/Shibboleth.sso
 # Be careful not to add a final /, as URLs like /Shibboleth.sso//Login will be generated,
 # which will cause Shibboleth errors.
 # To test using fakeshibboleth, use the value http://localhost:8001
 # To automatically use fakeshibboleth's .env file use http://localhost:8001/auto
-SHIBBOLETH_ROOT=http://localhost:8001
+SHIBBOLETH_ROOT=http://localhost:8001/auto
 
 # Whether we are running in production. This defines whether we have debug enabled, etc.
 UCLAPI_PRODUCTION=False
 
 # The domain we are running on. This is used by Django's ALLOWED_HOSTS feature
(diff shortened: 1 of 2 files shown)


backend/uclapi/.env.docker.example
```diff
@@ -35,7 +37,7 @@ WHITELISTED_CALLBACK_URLS=https://live-roombookings.uclapi.com/webhook/;https://
 DB_UCLAPI_NAME=uclapi
 DB_UCLAPI_USERNAME=uclapi
 DB_UCLAPI_PASSWORD=uclapi_test_password
-DB_UCLAPI_HOST=localhost
+DB_UCLAPI_HOST=db
```
greenfrogs, line 40: `.env.docker.example` is designed to be used with the `docker-compose-dev.yml` which exposes the database, gencache etc. to the host. These DNS names won't be able to be resolved (unless using the uclapi docker container).

backend/uclapi/.env.docker.example
```diff
@@ -2,29 +2,31 @@
...
-SECRET_KEY=
+SECRET_KEY=iRlsWf2BF2nnNy1zIsZcun6fZX5GGpk0
 # Root of Shibboleth's authentication. The example below might apply to fakeshibboleth;
 # an actual example would be https://uclapi.com/Shibboleth.sso
 # Be careful not to add a final /, as URLs like /Shibboleth.sso//Login will be generated,
 # which will cause Shibboleth errors.
 # To test using fakeshibboleth, use the value http://localhost:8001
 # To automatically use fakeshibboleth's .env file use http://localhost:8001/auto
-SHIBBOLETH_ROOT=http://localhost:8001
+SHIBBOLETH_ROOT=http://localhost:8001/auto
```
greenfrogs, line 12: Nice change 👍 this does cause quite a bit of confusion.

greenfrogs (changes requested):
Looks good but needs some changes to allow it to work with `docker-compose-dev.yml` (where uclapi and the frontend are built/developed outside of docker.

PR #3004, opened 2020-12-01
Add LibCal endpoints

Feverfew:
## What does this PR do?
Adds all of the LibCal endpoints that rely on the **Spaces - Read** scope
and the **Spaces - Write** scope.
Adds sphinx documentation for this new API.
However, website documentation has not been added.
Closes #2942
Closes #2943

## ✅ Pull Request checklist

- [x] Is this code complete?
- [x] Are tests done/modified?
- [ ] Are docs written for this PR? (if applicable)

## 🚨 Is this a breaking change for API clients?
No

## :squirrel: Deploy notes
`refresh_libcal_token` should be set to run every 30 mins, either via cron tab or systemd timer.

PR #3389, opened 2021-06-26
Add LibCal endpoints to OpenAPI

hueyy:
## What does this PR do?

This PR will add the new LibCal endpoints (#3004) to our OpenAPI config.

Added a new apiToken security scheme because an OAuth token is not actually needed to read non-personal LibCal data - a regular API token is suffifcient.

Non-personal read-only requests can be executed on the auto-generated OpenAPI docs, but not other requests because the OAuth security flow does not appear to be working.

## ✅ Pull Request checklist

- [x] Is this code complete?

No

## 🚀 Deploy notes

Rebuild frontend.

## Anything else

@shu8 we might want to split this 6k+ line JSON file into separate files in the future, perhaps make an openapi folder and separate JSON files for workspaces, timetable, LibCal, etc.

Not sure why codecov seems to think coverage in `uclapi/settings.py` has gone down.

### DIFF
--- a/uclapi.openapi.json
+++ b/uclapi.openapi.json
@@ -280,14 +280,58 @@
         "schema": {
           "type": "boolean"
         }
+      },
+      "details": {
+        "name": "details",
+        "in": "query",
+        "description": "Flag to indicate you want additional details such as terms and conditions.",
+        "required": false,
+        "example": 1,
+        "schema": {
+          "type": "integer"
+        }
+      },
+      "ids": {
+        "name": "ids",
+        "in": "query",
+        "description": "Comma delimited list of LibCal location IDs",
+        "required": true,
+        "example": "702,703",
+        "schema": {
+          "type": "string"
+        }
       }
     },
-    "schemas": {
-      "user_data": {
+    "schemas" : {
+      "libcal_location": {
         "properties": {
-          "ok": {
-            "type": "boolean",
-            "example": "true",
+          "lid": {
+            "type": "integer",
+            "example": 702,
+            "description": "ID for the LibCal location in question"
+          },
+          "name": {
+            "type": "string",
+            "example": "Bartlett Library",
+            "description": "Name of the LibCal location in question"
+          },
+          "public": {
+            "type": "integer",
+            "example": 1,
+            "description": "1 if the location is publicly-bookable, 0 if it is not"
+          },
+          "terms": {
+            "type": "string",
+            "example": "<strong>You must bring your UCL ID and a copy of the confirmation email to gain access to the space.</strong>",
+            "description": "Additional details about the location, such as terms & conditions. May include HTML."
+          }
+        }
+      },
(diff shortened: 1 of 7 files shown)


uclapi.openapi.json
```diff
@@ -272,9 +272,548 @@
...
+        "example": "cs_loYG4wTz,ab_P1bn30Mn",
+        "schema": {
+          "type": "string"
+        }
       }
     },
     "schemas" : {
+      "libcal_cancel_booking": {
+
+      },
```
shu8, line 440: is this supposed to be empty?

uclapi.openapi.json
```diff
@@ -272,9 +272,548 @@
...
+          "nickname": {
+            "type": "string",
+            "description": "If the space to be booked has a nickname, then the nickname must be supplied here.",
+            "example": "Tutorial Group Y"
+          },
+          "bookings": {
+            "type": "array",
+            "items": {
+
+            }
```
shu8, line 470: is this also supposed to be empty?

uclapi.openapi.json
```diff
@@ -272,9 +272,548 @@
...
+          },
+          "image": {
+            "type": "string",
+            "description": "A URL pointing to an image of the seat. If none is available, an empty string will be returned",
+            "example": "https://example.org/seat.png"
+          },
+          "status": {
+            "type": "boolean",
+            "description": "The current status of the seat",
+            "example": "Active"
```
shu8, line 568: this should be `true` if it's a bool

uclapi.openapi.json
```diff
@@ -4244,6 +4810,1679 @@
           }
         }
       }
+    },
+    "/libcal/space/locations": {
+      "get": {
+        "summary": "Gets all LibCal locations",
+        "tags": [ "LibCal" ],
```
shu8, line 4817: I think you might have forgotten to actually add this tag to the `tags` array (near the top of the file)!

uclapi.openapi.json
```diff
@@ -4244,6 +4810,1679 @@
...
+            "in": "query",
+            "description": "Comma delimited list of LibCal space IDs. If specified, will only show bookings for those spaces.",
+            "required": false,
+            "example": "469,470",
+            "schema": {
+              "type": "string"
+            }
+          },
+          {
+            "name": "seat_id",
```
shu8, line 5973: I think we use this multiple times so it might be better to move it into a `schema` and reference it whenever we need it?

Although I think at this point, this file is getting ridiculously big so this might be something to do after splitting the spec into multiple files -- see #3504

shu8 (changes requested):
Sorry for such a long delay :( This looks great, just a few comments!!

You're right -- this file is getting way too big. I've made an issue: #3504 and I'll look into this!

greenfrogs (approved):

shu8 (approved):

PR #3390, opened 2021-06-26
Log webhook execution and result

ILW8:
This adds last_successful, status_code fields to Webhook and WebhookTriggerHistory respectively. Allows to track how many webhooks are being run, and how many are successful. 

Needed to make a decision on  #3033 

## What does this PR do?
For example: Changes the way we retrieve bookings

## ✅ Pull Request checklist

- [x] Is this code complete?
- Are docs written for this PR? (Not applicable)

No

## 🚀 Deploy notes
Need to run migrations, adds two new columns to Webhook and WebhookTriggerHistory
Might need to run some tests, I'm not too sure on the datatype on L159.


backend/uclapi/roombookings/management/commands/trigger_webhooks.py
```diff
@@ -155,12 +152,23 @@ def webhook_filter(booking):
...
+                        webhook_success = True
+
                 webhook_in_db.last_fired = timezone.now()
                 webhook_in_db.save()
 
                 new_webhook_history_entry = WebhookTriggerHistory(
                     webhook=webhook_in_db,
-                    payload=webhook["payload"]
+                    payload=webhook["payload"],
+                    success=webhook_success,
```
greenfrogs, line 170: I don't feel there is any need to store a boolean for `success` as this is derivable from `status_code`

backend/uclapi/roombookings/management/commands/trigger_webhooks.py
```diff
@@ -155,12 +152,23 @@ def webhook_filter(booking):
         for webhook in webhooks_to_enact:
             if webhook["payload"]["content"] != {}:
                 webhook_in_db = webhook["webhook_in_db"]
+
+                webhook_success = False
+                status_code = -1
+                if "response" in webhook:
+                    status_code = webhook["response"].result().status_code
+                    if status_code not in range(400, 600):
```
greenfrogs, line 160: nit: change to < or >

ILW8:
why....

backend/uclapi/dashboard/migrations/0018_auto_20210625_0714_squashed_0021_auto_20210627_1527.py
```diff
@@ -0,0 +1,19 @@
...
+    dependencies = [
+        ('dashboard', '0017_webhook_last_success'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='webhook',
+            name='last_success',
+        ),
+    ]
```
shu8, line 19: this file seems to be removing the `last_success` field? should we only have the other migration that adds the field?

greenfrogs: https://docs.djangoproject.com/en/3.2/topics/migrations/#migration-squashing

greenfrogs:
> why....

Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.UndefinedColumn: column "last_success" of relation "dashboard_webhook" does not exist
LINE 1: ...rl", "siteid", "roomid", "contact", "last_fired", "last_succ...

backend/uclapi/dashboard/migrations/0018_auto_20210625_0714_squashed_0021_auto_20210627_1527.py
ILW8: I'm reading through the migrations and I see no mention of `WebhookTriggerHistory` model even before squashing, despite there being a change to that model. I squashed 18 to 21, then 17 to 18... but it resulted in nothing being applied in migrations...

ILW8:
Wooo it builds. Now to write tests...

ILW8:
I have no idea how mock.patch works with django's call_command. I couldn't get it to work with the way I usually use unittest.mock.patch. I've used quite a hacky workaround for now, would like to hear if there's a better way. 

Assuming that our guess was correct, these tests are going to be removed anyway once we remove webhooks support.

greenfrogs (approved):
Looks good to me

PR #3452, opened 2021-08-02
Migrate To url-loader

greenfrogs:
## What does this PR do?
Patches the controversial decision by webpack file loader to switch to using unflattened objects.

https://stackoverflow.com/questions/59070216/webpack-file-loader-outputs-object-module


ILW8 (approved):
LGTM

PR #3467, opened 2021-08-05
Enable OSSAR for repo

greenfrogs:
## What does this PR do?
Add https://github.com/github/ossar-action to run open source security scanning tools.


greenfrogs:
Probably not worth it as any errors are very likely to be picked up by CodeQL

PR #3574, opened 2021-09-13
docker cron: add resource limits for cron container

ILW8:
## What does this PR do?
Limit amount of resources the celery worker container can use. Should mitigate resource exhaustion.

## ✅ Pull Request checklist
- [x] Is this code complete?

No


greenfrogs:
I think we are just waiting for the OccupEye tasks (long, short, historical)

greenfrogs (approved):

PR #3738, opened 2021-11-24
backend common: add ping endpoint

greenfrogs:
## What does this PR do?
Adds a ping endpoint for use by Freshping, ?checks=1 can be added to run database, cache, and sentry checks.

## ✅ Pull Request checklist

- [X] Is this code complete?

No

### DIFF
--- a/backend/uclapi/common/views.py
+++ b/backend/uclapi/common/views.py
@@ -1,0 +1,67 @@
+
+from django.db import connections
+from django.core.cache import cache
+from common.helpers import PrettyJsonResponse
+
+from datetime import datetime, timedelta
+from time import sleep
+from uclapi.celery import ping_sample_task
+
+
+def ping_view(request):
+    if not request.GET.get('checks'):
+        return PrettyJsonResponse({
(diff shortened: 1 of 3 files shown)


backend/uclapi/common/views.py
```diff
@@ -0,0 +1,53 @@
...
+        cache.set("PING-TEST-CACHE", True, 30)
+        cache_connection = cache.get("PING-TEST-CACHE")
+    except Exception:
+        pass
+
+    celery_connection = False
+    start = time()
+    expires = datetime.now() + timedelta(seconds=10)
+    try:
+        task = ping_sample_task.apply_async(expires=expires)
```
ILW8, line 37: This would only test the generic worker, but not test the gencache worker

ILW8 (changes requested):

backend/uclapi/common/views.py
greenfrogs: Sorry that took so long, fixed now

ILW8 (approved):

PR #3739, opened 2021-11-27
backend celery: add occupeye cache to celery

ILW8:
## What does this PR do?
This completes the migration away from cronjobs, in favor of celery scheduled tasks.

## ✅ Pull Request checklist

- [x] Is this code complete?

No

### DIFF
--- a/backend/uclapi/workspaces/tasks.py
+++ b/backend/uclapi/workspaces/tasks.py
@@ -1,0 +1,53 @@
+from celery import shared_task
+from common.cachet import create_incident, CachetException, get_incident_name, update_incident
+from cachetclient.v1 import enums
+
+from workspaces.occupeye.cache import OccupeyeCache
+from workspaces.occupeye.endpoint import TestEndpoint
+
+
+@shared_task
+def feed_occupeye_cache(test=False, mini=False):
+    try:
+        print("Running OccupEye Caching Operation")
+        print("[+] Feeding Cache")
+        cache = OccupeyeCache(endpoint=TestEndpoint({})) if test else OccupeyeCache()
+        cache.feed_cache(full=(not mini))
+        print("Done!")
+        incident_name = get_incident_name("Occupeye")
+        if incident_name:
+            try:
+                update_incident("Occupeye Cache succeeded", incident_name, enums.INCIDENT_FIXED)
+            except CachetException as cachet_error:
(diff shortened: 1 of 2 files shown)


backend/uclapi/workspaces/tasks.py
```diff
@@ -0,0 +1,53 @@
...
+from common.cachet import create_incident, CachetException, get_incident_name, update_incident
+from cachetclient.v1 import enums
+
+from workspaces.occupeye.cache import OccupeyeCache
+from workspaces.occupeye.endpoint import TestEndpoint
+
+
+@shared_task
+def feed_occupeye_cache(test=False, mini=False):
+    try:
```
greenfrogs, line 11: can you please remove the try and except parts here

backend/uclapi/workspaces/tasks.py
```diff
@@ -0,0 +1,53 @@
...
+        print("Running OccupEye Caching Operation")
+        print("[+] Feeding Cache")
+        cache = OccupeyeCache(endpoint=TestEndpoint({})) if test else OccupeyeCache()
+        cache.feed_cache(full=(not mini))
+        print("Done!")
+        incident_name = get_incident_name("Occupeye")
```
greenfrogs, line 17: And the incident

backend/uclapi/workspaces/tasks.py
```diff
@@ -0,0 +1,53 @@
...
+        if incident_name:
+            try:
```
greenfrogs, line 19: And this try/catch here

backend/uclapi/workspaces/tasks.py
```diff
@@ -0,0 +1,53 @@
...
+
+@shared_task
+def day_cache():
+    feed_occupeye_cache.s(test=False, mini=True)()
+
+
+@shared_task
+def night_cache():
+    feed_occupeye_cache.s(test=False, mini=False)()
+    feed_occupeye_cache.s(test=False, mini=True)()
```
greenfrogs, line 53: I don't think we need to run a mini immediately. Also these might be able to be simplified and remove test=False (although not 100% sure how celery works with this)

greenfrogs (changes requested):
Overall it looks great but could you remove the try and except parts + incident + the try catch inside as Sentry would now catch any issues.

greenfrogs (approved):

PR #3811, opened 2022-02-10
libcal: use whitelist for sensitive endpoints instead of blacklist

shu8:
## What does this PR do?
Change GET bookings for libcal to only allow specific fields, rather than removing certain fields.

This means if the actual libcal endpoint adds any more sensitive fields in future, we won't return them to the user unwittingly.

## ✅ Pull Request checklist

- [x] Is this code complete?

No

### DIFF
--- a/backend/uclapi/libcal/tests.py
+++ b/backend/uclapi/libcal/tests.py
@@ -945,6 +945,27 @@ class LibcalPersonalEndpointsTestCase(APITestCase):
         self.assertJSONEqual(response.content.decode('utf8'), {"ok": True, 'bookings': cancelled})
 
     @parameterized.expand(['cs_qQpoVMHk', 'cs_qQpoVMHk,cs_qQpoRH20'])
+    def test_cancel_reject_other_user_booking(self, m, bookIds):
+        """Tests that only valid bookings belonging to the user are deleted."""
+        self.oauth_token.scope.scope_number = self.scopes_class.add_scope(0, 'libcal_write')
+        self.oauth_token.scope.save()
+        bookings = [{'email': 'not.this.user@example.com', 'bookId': 'cs_qQpoVMHk'}]
+
+        m.register_uri(
+            'GET',
+            f'https://library-calendars.ucl.ac.uk/1.1/space/booking/{bookIds}',
+            request_headers=self.headers,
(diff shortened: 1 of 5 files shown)


backend/uclapi/libcal/views.py
```diff
@@ -306,13 +306,11 @@ def get_bookings(request, *args, **kwargs):
...
-            for booking in data["bookings"]:
-                # Strip personal data from response
-                booking.pop('email', None)
-                booking.pop('first_name', None)
-                booking.pop('last_name', None)
-                booking.pop('book_id', None)
-                booking.pop('check_in_code', None)
+            for i in range(len(data["bookings"])):
+                # Ensure we don't leak personal data, only allow the following fields
+                allowed_fields = ['eid', 'cid', 'lid', 'from_date', 'to_date', 'created', 'status',
```
ILW8, line 311: We could perhaps have this be a settings entry, either in .env or in settings.py?

greenfrogs: I think this is probably okay, it is just for this endpoint and we don't want to pollute our .env or settings file.

backend/uclapi/libcal/tests.py
```diff
@@ -268,7 +271,11 @@ def test_bookings_personal_information_stripped(self, m):
...
             {'ids': 1, 'eid': 123, 'token': self.app.api_token}
         )
-        self.assertEqual(response.status_code, 200)
+
+        # Make sure the number of fields in the returned booking that are not in allowed_fields is 0
+        for booking in response.json()['bookings']:
+            self.assertEqual(len([x for x in booking.keys() if x not in allowed_fields]), 0)
+
         for k, v in response.json()['bookings'][0].items():
             self.assertEqual(k, 'leave_me')
```
greenfrogs, line 280: I don't believe this is tested anymore (as in the results is just empty)? Maybe add a seat_name or similar to the returned json?

greenfrogs (changes requested):

backend/uclapi/libcal/tests.py
shu8: ah thanks, updated!

greenfrogs (approved):

PR #3955, opened 2022-11-12
Don't accept OAuth tokens belonging to deleted apps

shu8:
## What does this PR do?
Rejects OAuth tokens if the corresponding app has been deleted.

## ✅ Pull Request checklist

- [x] Is this code complete?

Yes/No

## 🚀 Deploy notes

### DIFF
--- a/backend/uclapi/common/decorators.py
+++ b/backend/uclapi/common/decorators.py
@@ -160,6 +160,16 @@ def _check_oauth_token_issues(token_code, client_secret, required_scopes):
         response.status_code = 400
         return response
 
+    if token.app.deleted:
+        response = JsonResponse({
+            "ok": False,
+            "error":
+                "The token is invalid as the developer has "
+                "deleted their app."
+        })
+        response.status_code = 400
+        return response
+
     scopes = Scopes()
     for s in required_scopes:
         if not scopes.check_scope(token.scope.scope_number, s):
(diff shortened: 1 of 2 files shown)


backend/uclapi/common/decorators.py
```diff
@@ -160,6 +160,16 @@ def _check_oauth_token_issues(token_code, client_secret, required_scopes):
...
         return response
 
+    if token.app.deleted:
+        response = JsonResponse({
+            "ok": False,
+            "error":
+                "The token is invalid as the developer has "
+                "deleted their app."
+        })
+        response.status_code = 400
```
greenfrogs, line 170: nit: maybe a 403 error code instead? Or 401 if we are concerned about giving away information that this was a valid token (and change it so that we return 401 if token is invalid/missing).

greenfrogs (approved):
Happy, looks good.

backend/uclapi/common/decorators.py
greenfrogs: (just realised we give a specific reason so 403 is fine here)

shu8: good shout, done!

greenfrogs:
No need to update it on this PR but we have a commit style guide here: https://github.com/uclapi/uclapi/blob/master/CONTRIBUTING.md#git-commit-messages

PR #4059, opened 2023-04-12
Fix for Dockerfile smell DL3008

grosa1:
## What does this PR do?
Version pinning for installed apt packages

## ✅ Pull Request checklist

- [x] Is this code complete?

Yes

## 🚀 Deploy notes

## Anything else
Hi!
The Dockerfile placed at "docker/deployment/uclapi.Dockerfile" contains the best practice violation [DL3008](https://github.com/hadolint/hadolint/wiki/DL3008) detected by the [hadolint](https://github.com/hadolint/hadolint) tool.

The smell DL3008 occurs when the version pinning for the installed packages with apt is not specified. This could lead to unexpected behavior when building the Dockerfile.
In this pull request, we propose a fix for that smell generated by our fixing tool. We have verified that the patch is correct before opening the pull request.
To fix this smell, specifically, we use a heuristic approach that selects the most probable version tag for a given apt package corresponding to the latest version at the current date. The package versions are retrieved using the Canonical Launchpad APIs.

This change is only aimed at fixing that specific smell. If the fix is not valid or useful, please briefly indicate the reason and suggestions for possible improvements.

Thanks in advance

### DIFF
--- a/docker/deployment/uclapi.Dockerfile
+++ b/docker/deployment/uclapi.Dockerfile
@@ -25,26 +25,26 @@ WORKDIR /web
 COPY ./docker/deployment/non-public/${ENVIRONMENT}/uclapi/uclfw.rules /web/uclfw.rules
 
 RUN apt-get update && \
-    apt-get install -y python3 \
-                       python3-wheel \
-                       python3-setuptools \
-                       libaio1 \
-                       wget \
-                       git \
-                       libpq-dev \
-                       libpq5 \
-                       libpython3-dev \
-                       unzip \
-                       build-essential \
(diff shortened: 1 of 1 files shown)


zipy124:
This is not a smell in this case. This is intended behaviour as we base our images on the latest version of our chosen LTS release of ubuntu. Not specifying package versions allows for continous updates without specification on our end, and any errors are to be caught by the test suite.

I also note this seems to be a semi-automatically generated PR, please avoid submitting these, without looking at our specific project, and assessing if this is suitable for our case. Software development is not an absolute with rules to be followed to a tee, but guidelines.

PR #4161, opened 2023-11-21
Set room booking to return from current academic year

wilhelmklopp:
Just noticed that the room bookings API is still returning booking from the previous academic year and none for the current year. 

So for example:
`https://uclapi.com/roombookings/bookings?token=uclapi-abcd&start_datetime=2023-09-01`
returns:
```json
{
    "bookings": [],
    "next_page_exists": false,
    "count": 0,
    "ok": true
}
```

Whereas:
`https://uclapi.com/roombookings/bookings?token=uclapi-abcd&start_datetime=2023-04-01`
returns many bookings:
```json
{
    "bookings": [
        {
            "roomname": "C3.11",
            "siteid": "162",
            "roomid": "C3.11",
            "description": "Room Closed",
            "start_time": "2023-04-01T07:00:00+01:00",
            "end_time": "2023-04-01T21:00:00+01:00",
            "slotid": 6476822,
            "weeknumber": 31.0,
        },
        ...
    ],
    "next_page_exists": true,
    "page_token": "G30KYnWKki",
    "count": 121617,
    "ok": true
}
```

This PR should fix that (assuming the setid format is the same as it used to be)

Maybe in the future this code could be adjusted so that from say the 1st of September on any given year, UCL API automatically flips to the new academic year. And then no manual change is required each year.

## ✅ Pull Request checklist

- [X] Is this code complete?
- [X] Are docs written for this PR? (if applicable)

No

## 🚀 Deploy notes
Worth verifying this works as expected after deploying and after the regular rb dump job runs (if that's still how it works)

cc @zipy124 @shu8 

---

PS: I'm coming back to UCL for a guest lecture on the 29th of November at 2pm. Not yet sure where it'll be but would be cool to see you guys if you're free!  Guest lecture will be on using AI to help us build software and my experience with that.  Naturally, I'm planning one or more shout outs to UCL API 😎


wilhelmklopp:
@zipy124 thanks for merging ❤️ 

Does the change still need to be deployed?

I'm still getting the same empty response:

```
{
    "bookings": [],
    "next_page_exists": false,
    "count": 0,
    "ok": true
}
```

When querying for this: https://uclapi.com/roombookings/bookings?token=uclapi-abcd&start_datetime=2023-09-01

